PR #6, opened 2023-01-09
Latest contracts & Goerli deployment

justinkaseman:


KuphJr:
Can we remove the prettier PR for my own sanity?  We can do this later if it is necessary.

KuphJr:
I want to prioritize getting this merged so we can merge the `update-name` branch to main to avoid any further confusion on which branch to use.

KuphJr:
> > Can we remove the prettier PR for my own sanity? We can do this later in an isolated PR if it is necessary.
> 
> Sure if it makes it easier for you. You can also review by commit rather than all files changed.
> 
> We need a prettier config in this repo. I see it's added as a dependency, but since there is no config it is pulling from my global prettier config, which seems to be pretty different from your global config.

Thanks!  We should probably set some standard at some point, but that is certainly out of scope for the near future.

KuphJr:
I am still seeing a lot of changes in this PR replacing single-quotes w/ double quotes and other ancillary differences...

justinkaseman:
> I am still seeing a lot of changes in this PR replacing single-quotes w/ double quotes and other ancillary differences...

..this is the prettier thing that I just mentioned.
I can try to manually resolve, but it would be a lot easier if you just paste your prettier config so I can use the same.

justinkaseman:
> Can we eliminate everything except contract changes? The only changes in this PR outside of `/contracts` should the change from `FunctionsRegistry` to `FunctionsBillingRegistry`

There is also a bug fix or two in here and the new deployments.
Let's see how things look after the formatting consistency change and then if you really need it separate I guess I can do that.

tasks/Functions-admin/deployFullOracle.js
```diff
@@ -1,126 +1,124 @@
...
-      ? await oracle.setRegistry(registry.address, overrides)
-      : await oracle.setRegistry(registry.address)
-    console.log('Oracle registry set')
-  
-    if (process.env.ETHERSCAN_API_KEY || process.env.POLYGONSCAN_API_KEY) {
-      console.log('Waiting 6 blocks before verifying contracts...')
-      await setRegistryTx.wait(6)
+  }
+
+  console.log('Deploying Functions oracle factory')
```
justinkaseman, line 18: Oracle needs to deploy before registry because of allow list change for Automation

KuphJr: Good catch!

KuphJr (approved):

bolekk:
Can you rebase this on top of PR#10 please?

network-config.js
```diff
@@ -15,7 +15,7 @@ const networkConfig = {
                 process.env.MAINNET_RPC_URL ? '0xdc530d9457755926550b59e8eccdae7624181557' : ''
             }`,
         functionsPublicKey:
-            "971f006163a12ee3383a00d7743334480d6b1c83fdf60497e0c520b16d1a4ee421cc61375679b63466156fee6f2f1da5a7e630ba0b1cddb2704ef907ead223db",
+            "7e00a17e0d8c5c59bbe1f580f2405d51feb66c947fe66136190e80aabaf850964b837bcd379a92d5db52d2a8c8e044f8033cb981450bf3710ead0c4a43122ec1",
```
justinkaseman, line 18: @KuphJr I deployed with the secrets key pair that was in main previously. Do I need to switch it and re-deploy? :grimacing:

tasks/Functions-admin/whitelist.js
```diff
@@ -0,0 +1,20 @@
+const { networkConfig, VERIFICATION_BLOCK_CONFIRMATIONS } = require('../../network-config')
```
justinkaseman, line 1: @KuphJr Should this file be removed? Looks like a leftover from rebasing

KuphJr (approved):

bolekk (approved):
lgtm

PR #16, opened 2023-01-11
 fix ParseInt() bug in config file

zeuslawyer:
parseInt() can return NaN if the env var is set to a string.

This returns error when evaluated in the ternary operator.

Updated logic to check if the env var is a number and if it is

this will evaluate to undefined since this is an optional setting.


zeuslawyer:
Closing this as I see its already in the PR submitted by @KuphJr 
https://github.com/smartcontractkit/ocr2dr-hardhat-starter-kit/pull/15

PR #20, opened 2023-01-12
Added Sepolia network

pinebit:


.env.example
```diff
@@ -1,5 +1,7 @@
-MUMBAI_RPC_URL="example: https://rpc-mainnet.maticvigil.com      NOTE: set MUMBAI_RPC_URL or GOERLI_RPC_URL, but not both.  Delete one of these lines"
-GOERLI_RPC_URL="example: https://goerli.infura.io/v3/1234567890  NOTE: set MUMBAI_RPC_URL or GOERLI_RPC_URL, but not both.  Delete one of these lines"
+# NOTE: set only one RPC_URL, comment or delete others
+MUMBAI_RPC_URL="example: https://rpc-mainnet.maticvigil.com"
+GOERLI_RPC_URL="example: https://goerli.infura.io/v3/1234567890"
+SEPOLIA_RPC_URL="check infra for RPC url"
```
bolekk, line 4: We don't need all those examples, just use "[UNSET]" for all or some other marker.

hardhat.config.js
```diff
@@ -8,14 +8,20 @@ const MAINNET_RPC_URL = process.env.MAINNET_RPC_URL
...
-  )
+const selectedRPCs = [
+  MAINNET_RPC_URL, 
+  POLYGON_MAINNET_RPC_URL,
+  MUMBAI_RPC_URL,
+  GOERLI_RPC_URL,
+  SEPOLIA_RPC_URL
+].filter(v => !!v)
+if (selectedRPCs.length !== 1) {
+  throw Error("A single *_RPC_URL env variable must be set, found: " + selectedRPCs.join(","))
```
bolekk, line 22: I still don't get why we need to have exactly one of them set... It's more convenient to set them all. I'd even like to have per-network PRIVATE_KEYs set, so I can switch between networks easily.

.env.example
KuphJr: We should not set these values to anything if they are not needed.  Just leave them as `undefined`.

hardhat.config.js
KuphJr: Correct.  We no longer need only 1 set.  I fixed that.  The previous reason was to figure out which chain to fork for the simulator, but this was fixed by using the following statement:
`url: MAINNET_RPC_URL ?? POLYGON_MAINNET_RPC_URL ?? GOERLI_RPC_URL ?? MUMBAI_RPC_URL`

However, the statement above will not work correctly if we set the unset RPC URLs to `"UNSET"`

KuphJr (changes requested):

KuphJr:
Undefined URLs have been fixed in PR #21: https://github.com/smartcontractkit/ocr2dr-hardhat-starter-kit/pull/21

Please update this PR accordingly.

pinebit:
Updated

KuphJr:
I rebased it and made some minor changes to address the conflicts.  @pinebit I am going to approve and if everything looks good to you, go ahead & merge.

KuphJr (approved):

KuphJr (approved):

PR #26, opened 2023-01-14
Feat/off chain secrets encrypter

KuphJr:


Functions-request-config.js
```diff
@@ -26,7 +27,7 @@ const requestConfig = {
   // location of source code (only Inline is curently supported)
   codeLocation: Location.Inline,
   // location of secrets (only Inline is currently supported)
-  secretsLocation: Location.Inline,
+  secretsLocation: Location.Offchain,
```
bolekk, line 30: We don't want this to be a default for our example request. The example should work out of the box.

Functions-request-config.js
```diff
@@ -26,7 +27,7 @@ const requestConfig = {
   // location of source code (only Inline is curently supported)
   codeLocation: Location.Inline,
   // location of secrets (only Inline is currently supported)
```
bolekk, line 29: Fix comment.

Functions-request-config.js
```diff
@@ -47,6 +48,18 @@ const requestConfig = {
   maxResponseBytes: 256,
   // expected type of the returned value
   expectedReturnType: ReturnType.uint256,
+  // Reduntant URLs which point to encrypted off-chain secrets
+  // TODO: The tooling should fetch this URL & verify that the secrets are correctly specified (ie: has the correct format & contains entries for all current transmitter node addresses)
+  secretsURLs: [
+    'https://raw.githubusercontent.com/KuphJr/off-chain-secrets/main/test-secrets.json'
+  ],
+  // Per-node offchain secrets objects used by the `functions-build-offchain-secrets` command
```
bolekk, line 56: Again, the example should use just a single key.

FunctionsRequestSimulator/Functions.js
```diff
@@ -1,145 +1,155 @@
...
-    return mod && mod.__esModule ? mod : { default: mod }
-  }
-var _a
-Object.defineProperty(exports, "__esModule", { value: true })
-exports.FunctionsModule = void 0
-const axios_1 = __importDefault(require("axios"))
-const Log_1 = require("./Log")
+"use strict";
+var __importDefault = (this && this.__importDefault) || function (mod) {
+    return (mod && mod.__esModule) ? mod : { "default": mod };
```
bolekk, line 3: You need to separate unrelated formatting from new logic. Otherwise this PR is too hard to review. Almost every code line in the repo has been changed ...

bolekk (changes requested):
Please separate formatting changes from new logic.

FunctionsRequestSimulator/Functions.js
KuphJr: The changed code is not a formatting change.  This is because the files you see in the `./FunctionsRequestSimulator` folder are transpiled JavaScript files from Typescript.  The original TypeScript changes can be found in this PR: https://github.com/smartcontractkit/universal-adapter-sandbox/pull/14

Eventually this should be an NPM package, but creating an NPM package from the sandbox simulator is currently out of scope.

In order for the code in the PR to work properly, it must include the changes in the `./FunctionsRequestSimulator` directory in order to build the off-chain secrets.

bolekk: The fundamental problem is still the same. Can you please separate all auto-generated changes into a separate PR (with minimal other changes to keep the simulator working)? Then on top of it, send a PR that implements off-chain secret encryption in tooling. Also if it depends on a PR in another repo, we should get that other one merged first.

Please also add a description of what else is changing. For example are you switching the default example to use off-chain secrets? - that was one of my comments as well.

PR #27, opened 2023-01-14
Provide option to verify deployed contract. Don't verify based on the existence of some environment variables

### LINKED ISSUE #23: Verification is not generic enough
https://github.com/smartcontractkit/ocr2dr-hardhat-starter-kit/blob/d995299bd0c0da492629872995f1a622a4bb7d8c/tasks/Functions-client/deploy.js#L25

Hi @justinkaseman @KuphJr 

this line is not generic enough. Each time that a new network is supported, the line will have to be modified. I'd propose something like this:

- Add an optional task parameter to ask the user if they want to verify their contract (e.g.:  --verify true)
- Rely on the standard verify plugin mechanism. The verify task would fail if the user didn't provide the right API_KEY. Because you catch the error, you will display the reason for verification failure at the end

aelmanaa:
fix #23


pinebit (approved):

bolekk:
Please rebase.

aelmanaa:
@bolekk rebase done.

PR #31, opened 2023-01-16
document numAllowedQueries in README

aelmanaa:


README.md
```diff
@@ -86,6 +86,7 @@ Chainlink Functions requests can be configured by modifying values in the `reque
 | `secretsLocation`    | This specifies where the encrypted secrets for a request are located. Currenly, only the `Inline` option is supported (represented by the value `0`). This means encypted secrets are provided directly in the on-chain request instead of being referenced via a URL or IPFS hash.  |
 | `codeLanguage`       | This specifies the language of the source code which is executed in a request. Currently, only `JavaScript` is supported (represented by the value `0`).                                                                                                                             |
 | `source`             | This is a string containing the source code which is executed in a request. This must be valid JavaScript code that returns a Buffer. See the [JavaScript Code](#javascript-code) section for more details.                                                                          |
+| `numAllowedQueries`             | This is a number defining the maximum number of HTTP queries the source code is allowed to make.                                                                          |
```
bolekk, line 89: I think that `numAllowedQueries` an `maxResponseBytes` should be removed from our examples and docs altogether. We cap those values for security reasons anyway, so beyond some point user's settings won't matter and will be confusing. cc @KuphJr Morgan how about we remove them?

bolekk (changes requested):

README.md
KuphJr: I agree.  These parameters and currently set by the external adapter and are not configurable by the user.  We may decide to add user-configurable limits w/ different costs in the future, but that is well out of scope for the time being.

I will ensure all references to `numAllowedQueries` and `maxResponseBytes` are removed.

PR #33, opened 2023-01-16
Open api sample

aelmanaa:
- Add openApi (no authentication) example
- Allow providing a path to external request config file to simulate, request, and read-latest tasks. the path parameter is optional. One can provide request config file as follows:
   -  simulate: `npx hardhat functions-simulate --path ./samples/open-api/Functions-request-config.js`
   -  execute: `npx hardhat functions-request --network NETWORK --contract CLIENT_CONTRACT --subid SUBSCRIPTION_ID --path ./samples/open-api/Functions-request-config.js`


samples/common.js
```diff
@@ -0,0 +1,23 @@
+const Location = {
```
bolekk, line 1: We already have all these definitions in the top-level Functions-request-config.js, let's not repeat them again but create a shared, top-level common file.

tasks/Functions-client/simulate.js
```diff
@@ -49,7 +51,16 @@ task("functions-simulate", "Simulates an end-to-end fulfillment locally for the
     await registry.addConsumer(subscriptionId, client.address)
 
     // Build the parameters to make a request from the client contract
-    const requestConfig = require("../../Functions-request-config.js")
+    const requestConfigPath = taskArgs.path
```
bolekk, line 54: This is again a copy-pasted block, let's share code.

bolekk (changes requested):

aelmanaa:
@bolekk  Enhanced the scripts as well. the Request config file is not hardcoded so you can provide your owner config:

- If you provide your own file: `npm run functions-build-request -- --path ./samples/open-api/Functions-request-config.js` 
- If you want to use the default file: `npm run functions-build-request`

KuphJr:
I am concerned about growing the complexity of this repo.  I prefer there to be as few files as possible that a user must grok.
This repo should be the minimum example a developer needs to make their first request and understand the available features.
Additionally, it should demonstrate good practices such as using multiple APIs whenever possible to avoid centralization (hence the multiple API example that demonstrates using both open APIs and an API that requires authentication.)

Eventually, our goal is to provide users with a HardHat plugin that they can import when they become more accustomed to using this product.  I think we should reserve further complexity for the plugin, but leave this repo as-is.

PR #38, opened 2023-01-18
Specify in the quickstart how to verify your client contract

aelmanaa:


KuphJr:
This needs a note that it will only work if they supply an Etherscan or Polygonscan API key, or add a check in the `deploy.js` file that those values are set and give a warning message if they are not.

aelmanaa:
@KuphJr Note added

KuphJr (approved):

PR #39, opened 2023-01-18
fix OCRReponse event in simulation

aelmanaa:


KuphJr (approved):

KuphJr:
Thanks for this catch!  I had fixed this issue in `request.js` but not `simulate.js`

PR #40, opened 2023-01-18
Typo fixes and suggested changes to readme content

PickleyD:


README.md
```diff
@@ -15,9 +15,9 @@
 
 # Overview
 
-<p>Chainlink Functions allows users to request data from almost any API and perform custom computation using JavaScript.</p>
-<p>It works by using a <a href="https://chain.link/education/blockchain-oracles#decentralized-oracles">decentralized oracle network</a> (DON).<br>When a request is initiated, each node in the DON executes the user-provided JavaScript code simultaneously.  Then, nodes use the <a href="https://docs.chain.link/architecture-overview/off-chain-reporting/">Chainlink OCR</a> protocol to come to consensus on the results.  Finally, the median result is returned to the requesting contract via a callback function.</p>
-<p>Chainlink Functions also enables users to share encrypted secrets with each node in the DON.  This allows users to access to APIs that require authentication, without exposing their API keys to the general public.
+<p>Chainlink Functions allows smart contracts to perform custom computation off-chain in a decentralized manner.</p>
```
PickleyD, line 18: Suggested change here because the first line didn't mention smart contracts and I don't think JavaScript needs to be explicitly mentioned here if it's only the first language to be supported

KuphJr: Thanks for the suggestion!

Right now I think the existing description of the utilization of the DON is important, so I am going to close this PR, but I will make a note to adjust the README to remove the reference to JavaScript in the future.

PR #44, opened 2023-01-19
re-added deleted FunctionsOracleConfig.json file

KuphJr:


pinebit (approved):

bolekk:
This file is network-specific and it doesn't make sense to keep it here (should be re-generated on every contract update). There's also no reason for the simulator to depend on it so let's change that.

PR #58, opened 2023-01-25
user-facing offchain secrets tooling

KuphJr:
- Updated contracts with those in `core` (only change is to add ability to use `Remote` secrets)
- Added `functions-build-offchain-secrets` command
- Added `functions-build-request-JSON` command (This is used to generate all required request parameters, including encrypted secrets or secrets URLs, which can be copy-pasted into Remix if desired)
- Added ability to use offchain secrets in `functions-request`
- Added ability to use offchain secrets in `functions-simulate`
- The `npm run functions-simulate-javascript` command now runs a simulation for each set of per-node secrets to ensure validity

- TODO: Update readme to include documentation for off-chain secrets:
  - Goal: add a separate section describing how to use off-chain secrets (This work can be included in a separate PR)


contracts/dev/functions/Functions.sol
```diff
@@ -116,6 +116,18 @@ library Functions {
     self.secrets = secrets;
   }
 
+  /**
+   * @notice Adds Remote user encrypted secrets to a Request
+   * @param self The initialized request
+   * @param encryptedSecretsURLs Encrypted comma-separated string of URLs pointing to off-chain secrets
+   */
+  function addRemoteSecrets(Request memory self, bytes memory encryptedSecretsURLs) internal pure {
```
bolekk, line 124: It would be better to get it out of the way in a separate PR.

Functions-request-config.js
```diff
@@ -40,6 +41,11 @@ const requestConfig = {
   args: ["1", "bitcoin", "btc-bitcoin", "btc", "1000000", "450"],
   // expected type of the returned value
   expectedReturnType: ReturnType.uint256,
+  // Redundant URLs which point to encrypted off-chain secrets
+  secretsURLs: [],
+  // Per-node offchain secrets objects used by the `functions-build-offchain-secrets` command
+  // The first entry will be used by the simulator if `secrets` is undefined
```
bolekk, line 47: I think we need a better explanation. Something like "instead of a single set of secrets sent to all nodes, you can specify per-node secrets. To do so, leave 'secrets' undefined and define this map. The keys are .."

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+  
+        const secretsObjectKeys = JSON.stringify(Object.keys(secrets).sort())
+  
+        if (
+          requestConfig.secrets
+          && defaultSecretsObjectKeys.length > 0
+          && defaultSecretsObjectKeys !== secretsObjectKeys
+        ) {
+          throw Error('In the config file, not all objects in `perNodeSecrets` have the same object keys as default `secrets`. (The values can be different, but the keys should be the same between all objects.)')
```
bolekk, line 58: I wonder if we could relax that to allow per-node secrets for one auth key and a global default for another. Not important right now

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+  
+        for (const comparedSecrets of requestConfig.perNodeSecrets) {
```
bolekk, line 61: No need for n^2 comparisons, just compare only to the first one.

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -0,0 +1,115 @@
...
+
+    const [ nodeAddresses, publicKeys ] = await oracleContract.getAllNodePublicKeys()
+
+    if (
+      requestConfig.perNodeSecrets
+      && requestConfig.perNodeSecrets.length !== nodeAddresses.length
+      && requestConfig.perNodeSecrets.length !== 0
+    ) {
+      throw Error(
+        `The number of per-node secrets must match the number of nodes.  Length of perNodeSecrets: ${requestConfig.perNodeSecrets.length} Number of nodes: ${nodeAddresses.length}`
```
bolekk, line 81: Can we do this check first?

tasks/Functions-client/request.js
```diff
@@ -230,3 +245,41 @@ task("functions-request", "Initiates a request from an Functions client contract
       console.log(`Waiting for fulfillment...\n`)
     })
   })
+
+
+const verifyOffchainSecrets = async (secretsURLs, nodeAddresses) => {
```
bolekk, line 250: Reuse code instead of copy-pasting.

tasks/Functions-client/buildRequestJSON.js
```diff
@@ -0,0 +1,99 @@
+const { simulateRequest, buildRequest, getRequestConfig } = require("../../FunctionsSandboxLibrary")
+const { networkConfig } = require("../../network-config")
+const readline = require("readline-promise").default
+const axios = require("axios")
+const fs = require("fs")
+
+task("functions-build-request", "Creates a JSON file with Functions request parameters")
```
bolekk, line 7: Is this necessary to have? It might be confusing to have both this and our tooling having a send-request command. Also it could go in a separate PR as it doesn't seem to be required for offchain secrets, right?

bolekk (changes requested):
Initial pass.

Functions-request-config.js
```diff
@@ -5,6 +5,7 @@ require("dotenv").config()
 
 const Location = {
   Inline: 0,
+  Remote: 1,
```
pinebit, line 8: The plugin project uses https://github.com/dethcrypto/TypeChain to auto-generate all typings from Solidity code. This is amazing property and benefit of using TypeScript (someday). FYI.

tasks/Functions-client/buildRequestJSON.js
KuphJr: I do think it is useful.  I personally use it a lot and if users want to use some alternate tool for deploying and interacting with contracts, it is also useful.

However, I would support keeping this feature (in a separate PR), just because it is something I personally use frequently.

Functions-request-config.js
KuphJr: I want to leave the more in-depth explanation for the README and have the inline comments be short & sweet.

Do you still want me to modify the comment or can I just include this in the README?

tasks/Functions-client/request.js
KuphJr: The `verifyOffchainSecrets` command is not used anywhere else except `request.js`.

It is placed in a separate function to avoid further confusing the existing logic in `request.js`.

bolekk: It is used in buildRequestJSON.js - are you planning to share that one?

KuphJr: No.  `verifyOffchainSecrets` checks the URLs where `buildRequestJSON.js` builds the JSON file which must be uploaded.  I will rename the function for clarity.

bolekk: My point is that there's literally the same piece of code in two PRs so why not put it in a common place to reuse it?

KuphJr: I see where I missed the repeated use.  It was because I had already moved buildRequestJSON into a different PR.

KuphJr: Since these 2 PRs have been split, this must be addressed after the PRs are merged (see my comments in the `functions-build-request` PR)

Functions-request-config.js
bolekk: That's fine.

tasks/Functions-client/request.js
bolekk: Let's merge PR #59 and then rebase this one and address it here.

KuphJr:
@bolekk All feedback has been addressed.  It should be ready for final review before tomorrow's demo!

I appreciate all the input! 👍

scripts/simulateFunctionsJavaScript.js
```diff
@@ -1,12 +1,37 @@
...
   }
+}
+
+(async () => {
+    const unvalidatedRequestConfig = require("../Functions-request-config.js")
+    const requestConfig = getRequestConfig(unvalidatedRequestConfig)
+
+    if (requestConfig.secretsLocation === 1) {
+      if (requestConfig.secrets && Object.keys(requestConfig.secrets).length !== 0) {
+        console.log('\n__SIMULATING JAVASCRIPT WITH DEFAULT SECRETS__'.toUpperCase())
```
pinebit, line 19: what is the purpose of uppercasing an uppercase string?

scripts/simulateFunctionsJavaScript.js
```diff
@@ -1,12 +1,37 @@
...
+
+      for (let i = 0; i < requestConfig.perNodeSecrets.length; i++) {
+        requestConfig.secrets = requestConfig.perNodeSecrets[i]
+        console.log(`\n__SIMULATING JAVASCRIPT WITH SECRETS ASSIGNED TO NODE ${i}__`.toUpperCase())
+        await runSimulation(requestConfig)
+      }
+      return
+    }
+
+    console.log('\n__Simulating JavaScript__')
```
pinebit, line 35: And this is not uppercase.. I am wondering why some log messages are uppercase while some are not?
I think all tooling output shall be consistent.

pinebit (commented):
Minor comments

scripts/simulateFunctionsJavaScript.js
```diff
@@ -1,12 +1,37 @@
...
+    console.log(`Value returned from source code: ${result}\n${getDecodedResultLog(requestConfig, result)}`)
     return
   }
+}
+
+(async () => {
+    const unvalidatedRequestConfig = require("../Functions-request-config.js")
+    const requestConfig = getRequestConfig(unvalidatedRequestConfig)
+
+    if (requestConfig.secretsLocation === 1) {
```
bolekk, line 17: replace "1" with "Remote"

scripts/simulateFunctionsJavaScript.js
```diff
@@ -1,12 +1,37 @@
...
+      if (requestConfig.secrets && Object.keys(requestConfig.secrets).length !== 0) {
+        console.log('\n__SIMULATING JAVASCRIPT WITH DEFAULT SECRETS__'.toUpperCase())
+        await runSimulation(requestConfig)
+      }
+
+      if (!requestConfig.perNodeSecrets) {
+        return
+      }
+
+      for (let i = 0; i < requestConfig.perNodeSecrets.length; i++) {
```
bolekk, line 27: I feel like this should be under a flag. Not everybody wants to repeat a simulation N times every single time.

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        if (firstPerNodeSecretsKeys !== secretsObjectKeys) {
+          throw Error('In the config file, not all objects in `perNodeSecrets` have the same object keys. (The values can be different, but the keys should be the same between all objects.)')
+        }
+      }
+    }
+
+    console.log(`Using public keys from FunctionsOracle contract ${networkConfig[network.name]['functionsOracle']} on network ${network.name}`)
+    const OracleContract = await ethers.getContractFactory("FunctionsOracle")
```
bolekk, line 70: This should be called *Factory. It's too confusing to have two vars that differ only by casing. It's JS after all, not Go ;)

tasks/Functions-client/request.js
```diff
@@ -44,7 +45,20 @@ task("functions-request", "Initiates a request from an Functions client contract
     const registry = await RegistryFactory.attach(registryAddress)
 
     console.log("Simulating Functions request locally...")
-    const requestConfig = require("../../Functions-request-config.js")
+    const unvalidatedRequestConfig = require("../../Functions-request-config.js")
+    const requestConfig = getRequestConfig(unvalidatedRequestConfig)
+
+    if (requestConfig.secretsLocation === 1) {
```
bolekk, line 51: Replace "1" with the enum

tasks/Functions-client/simulate.js
```diff
@@ -72,7 +72,17 @@ task("functions-simulate", "Simulates an end-to-end fulfillment locally for the
 
       // Simulating the JavaScript code locally
       console.log("\nExecuting JavaScript request source code locally...")
-      const { success, result, resultLog } = await simulateRequest(require("../../Functions-request-config.js"))
+      const unvalidatedRequestConfig = require("../../Functions-request-config.js")
+      const requestConfig = getRequestConfig(unvalidatedRequestConfig)
+  
+      if (requestConfig.secretsLocation === 1) {
```
bolekk, line 78: Same here

bolekk (changes requested):
Left a few comments but overall looks OK

scripts/simulateFunctionsJavaScript.js
KuphJr: It was to easily identify which log statements belong to each each per-node execution.

scripts/simulateFunctionsJavaScript.js
KuphJr: The problem is there is no consistent cross-platform way to set flags for `npm` commands.

Maybe I'll just let a user set an env var if they want to turn this off.

tasks/Functions-client/buildOffchainSecrets.js
KuphJr: lol its also confusing that one of our contracts is called OracleFactory too.  However I will make the change here.

scripts/simulateFunctionsJavaScript.js
KuphJr: I personally find it useful, so I would advocate to keep it.  It is not a function that can be included in the HardHat plugin anyway since it is just an NPM command so a user can run a sanity check on their code.

scripts/simulateFunctionsJavaScript.js
KuphJr: Else, it is going to be a larger change to make Functions-request-config.js have separate exports as opposed to a default export.

KuphJr: So, its whatever we prefer.  Either repeat code for the Location object "enum", create a new utils file that will hold `Location`, or leave it as is.

KuphJr: IMO, best solution is to address this in a separate PR that introduces a `utils` file that also holds things like the gas overrides.

KuphJr: This "enum" object will likely go away entirely when we have proper TypeScript implementation for the plugin.

bolekk: Yes, let's move to common.js in a separate PR.

bolekk (approved):

PR #59, opened 2023-01-25
added functions-build-request command

KuphJr:
- Added `functions-build-request` command which outputs a JSON file with required request parameters `source`, `args` & encrypted `secrets` from `Functions-request-config.js`
  - This can be used to build request parameters which can be entered into another tool such as Remix or Tenderly
  - This is also useful for building test cases


pinebit (approved):

tasks/Functions-client/buildRequestJSON.js
```diff
@@ -0,0 +1,99 @@
...
+        url,
+        secrets: response.data,
+      })
+    } catch (error) {
+      throw Error(`Failed to fetch off-chain secrets from ${url}\n${error}`)
+    }
+  }
+
+  for (const { secrets, url } of offchainSecretsResponses) {
+    for (const comparedSecretsResponse of offchainSecretsResponses) {
```
bolekk, line 84: I left a comment on the other PR that there's no need for nested loops here.

tasks/Functions-client/buildRequestJSON.js
```diff
@@ -0,0 +1,99 @@
...
+    requestConfig.DONPublicKey = DONPublicKey.slice(2)
+    // Build the parameters to make a request from the client contract
+    const request = await buildRequest(requestConfig)
+
+    fs.writeFileSync(taskArgs.output ?? 'Functions-request.json', JSON.stringify(request))
+    console.log(`Wrote request to ${taskArgs.output ?? 'Functions-request.json'}`)
+  })
+
+
+const verifyOffchainSecrets = async (secretsURLs, nodeAddresses) => {
```
bolekk, line 64: Are you planning to call this one from request.js?

bolekk (changes requested):

tasks/Functions-client/buildRequestJSON.js
KuphJr: Yes, however this must be addressed after the other PR is merged since both PRs require this logic but they have now been split into separate PRs.

KuphJr: Realistically, a lot of re-used logic can be put into a common `utils` to prevent repetition (such as the gas limit overrides), but this should be a separate PR.

bolekk (approved):

PR #62, opened 2023-01-26
Use @chainlink/contracts package for dev contracts

justinkaseman:


pinebit:
Why do we need to keep `contracts/dev/FunctionsOracle.sol` and friends? and not just reference those from the package?

justinkaseman:
> Why do we need to keep `contracts/dev/FunctionsOracle.sol` and friends? and not just reference those from the package?

So that hardhat compiles them and holds the artifacts (ABIs) for them.
See https://ethereum.stackexchange.com/questions/114376/how-to-compile-external-contracts-using-hardhat

KuphJr:
> > Why do we need to keep `contracts/dev/FunctionsOracle.sol` and friends? and not just reference those from the package?
> 
> So that hardhat compiles them and holds the artifacts (ABIs) for them that we use within the CLI commands. See https://ethereum.stackexchange.com/questions/114376/how-to-compile-external-contracts-using-hardhat



> > Why do we need to keep `contracts/dev/FunctionsOracle.sol` and friends? and not just reference those from the package?
> 
> So that hardhat compiles them and holds the artifacts (ABIs) for them that we use within the CLI commands. See https://ethereum.stackexchange.com/questions/114376/how-to-compile-external-contracts-using-hardhat



> > Why do we need to keep `contracts/dev/FunctionsOracle.sol` and friends? and not just reference those from the package?
> 
> So that hardhat compiles them and holds the artifacts (ABIs) for them that we use within the CLI commands. See https://ethereum.stackexchange.com/questions/114376/how-to-compile-external-contracts-using-hardhat

Just so I understand, is `FunctionsOracle.sol` is not included in our imported package?  Is that why it still must be provided here?

network-config.js
```diff
@@ -43,10 +43,10 @@ const networkConfig = {
   },
   sepolia: {
     linkToken: "0x779877A7B0D9E8603169DdbD7836e478b4624789",
-    linkEthPriceFeed: "",
-    functionsOracle: "0x88F9Dd856D2c8CeC68569b57858ee953CC2bB4f4",
-    functionsOracleFactory: "0x85b44e2A2812E0907136357bb5F536da7cC99c4A",
-    functionsOracleRegistry: "0x774Ea40eCC9fBC7861b35d57B5f49E13Ce87d75C",
+    linkEthPriceFeed: "0x42585eD362B3f1BCa95c640FdFf35Ef899212734",
+    functionsOracle: "0x0019f8318006EB71594c4a6CF3cfcf1B25b2174b",
```
bolekk, line 47: This needs to happen separately, when we update the DON.

bolekk (changes requested):
Once this is merged, it will break current DONs, since they're using older contract versions. We need to coordinate a re-deployment of everything.

network-config.js
pinebit: I do agree this change is unrelated to the topic of this PR. Let's revert this part and everything else looks good.

justinkaseman:
Made obsolete by https://github.com/smartcontractkit/functions-hardhat-starter-kit/pull/73

PR #64, opened 2023-01-27
Rework cost estimation gas prices

justinkaseman:
TODO:
- [ ] Rework mumbai gas cost estimation
- [ ] Add simulator estimations to `request` command

### DIFF
--- a/tasks/Functions-client/request.js
+++ b/tasks/Functions-client/request.js
@@ -1,4 +1,9 @@
-const { simulateRequest, buildRequest, getDecodedResultLog, getRequestConfig } = require("../../FunctionsSandboxLibrary")
+const {
+  simulateRequest,
+  buildRequest,
+  getDecodedResultLog,
+  getRequestConfig,
+} = require("../../FunctionsSandboxLibrary")
 const { VERIFICATION_BLOCK_CONFIRMATIONS, networkConfig } = require("../../network-config")
 const { verifyOffchainSecrets } = require("./buildRequestJSON")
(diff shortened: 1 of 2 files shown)


tasks/Functions-client/request.js
```diff
@@ -101,7 +101,14 @@ task("functions-request", "Initiates a request from an Functions client contract
     const request = await buildRequest(requestConfig)
 
     // Estimate the cost of the request
-    const { lastBaseFeePerGas, maxPriorityFeePerGas } = await hre.ethers.provider.getFeeData()
+    let gasPriceEstimate
+    const { lastBaseFeePerGas, maxPriorityFeePerGas, gasPrice } = await hre.ethers.provider.getFeeData()
+    if (network.name === "mumbai") {
+      gasPriceEstimate = gasPrice.mul(lastBaseFeePerGas)
```
bolekk, line 107: wow, is that what we're supposed to do on Mumbai? Can you add a comment explaining why + a source link if any?

KuphJr: I am also curious as well.  Why does Mumbai require this different estimation?

KuphJr (approved):
Approving now as it appears to work just fine, but I am still curious for an explanation of the difference between networks

bolekk (changes requested):
This approximation can break any time. Let's not merge it.

bolekk:
We should debug and fix it properly, not use an arbitrary approximation.

PR #65, opened 2023-01-27
UX additions

justinkaseman:


KuphJr:
Nice rewording! It does make it easier to understand.  Also thanks for the whitelist check.  I am sure that will save us for a multitude of questions.

KuphJr (approved):

PR #68, opened 2023-01-27
Basic CI/CD

justinkaseman:
### Description
Adds:
1. Basic Github actions for running tests & linting
2. Placeholder unit test

Fixes:

1. Husky was not running on pre-commit, which is what was causing prettier to never format as it does in other CLL repos


KuphJr:
I don't envision us needing the contract unit tests as our contract really isn't much more than a template, however maybe we can keep it in case that changes in the future.

justinkaseman:
> I don't envision us needing the contract unit tests as our contract really isn't much more than a template, however maybe we can keep it in case that changes in the future.

Isn't the idea that users will use this repo to develop their dApp?
Or what did you envision as the user workflow, they copy their contracts in from another repository?

KuphJr:
I want to make sure this works before I approve and merge.  I currently see that the checks have been queued and pending for 10 minutes...

I also don't want to get too bogged down on debugging this if it does prove difficult as we have more pressing matters prior to launch.

KuphJr:
> > I don't envision us needing the contract unit tests as our contract really isn't much more than a template, however maybe we can keep it in case that changes in the future.
> 
> Isn't the idea that users will use this repo as a starter to develop their dApp? Or what did you envision as the user workflow, they copy their contracts in from another repository, deploy, and then go back to their own solidity repo?

I was just referring to our workflow as I don't expect us to be adding or modifying the contracts in this repo internally.  However, maybe Dev Rel or another team developing examples might in the future?  IDK, but seeing as it is a stub test I see no reason to remove it either.

justinkaseman:
> I want to make sure this works before I approve and merge. I currently see that the checks have been queued and pending for 10 minutes...
> 
> I also don't want to get too bogged down on debugging this if it does prove difficult as we have more pressing matters prior to launch.

I think it's because it is trying to use CLL runners, but they aren't configured. I'll just switch to Github's public runners for the time being.

KuphJr:
Note from Slack messages for visibility:

> KuphJr: I think I recall having issues with Husky pre-commit on my windows desktop a year ago when I was doing the take-home project.  (although it might have been something else).  I recall having to do the project on my old Macbook.

This needs to be tested on all platforms to ensure beginning devs do not encounter any issues or error messages due to our Husky pre-commit.  Furthermore, this needs to be tested on Windows.

We need to 100% confirm our workflows do not impact the end users of this repo in any way before this is merged.

Until then, we all must do our own due diligence to run Prettier before making PRs

KuphJr (approved):

KuphJr:
I am performing some testing of the entire tooling suite on windows and have not yet noticed a problem.

I will merge this myself once the testing concludes.

KuphJr (approved):

PR #69, opened 2023-01-27
globalSecrets

KuphJr:
This PR uses the new `globalSecrets` parameter instead of `secrets` for off-chain secrets.


Functions-request-config.js
```diff
@@ -43,8 +43,9 @@ const requestConfig = {
   expectedReturnType: ReturnType.uint256,
   // Redundant URLs which point to encrypted off-chain secrets
   secretsURLs: [],
+  // Default offchain secrets object used by the `functions-build-offchain-secrets` command
+  globalSecrets: {},
```
bolekk, line 47: Can we remove "secrets" from this file, so as not to confuse users?

README.md
```diff
@@ -85,18 +85,19 @@ Example: `npx hardhat functions-read --network goerli --contract 0x787Fe00416140
...
+| Setting Name         | Description                                                                                                                                                                                                                                                                                                            |
+| -------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `codeLocation`       | This specifies where the JavaScript code for a request is located. Currently, only the `Location.Inline` option is supported (represented by the value `0`). This means the JavaScript string is provided directly in the on-chain request instead of being referenced via a URL.                                      |
```
bolekk, line 90: Why does this section keep changing in every PR?

Functions-request-config.js
KuphJr: secrets is now user for on-chain secrets.  I thought it actually made more sense to do it this way and let to less confusion overall, but let me know if you feel otherwise.

KuphJr: I could call it `globalOffchainSecrets` and `perNodeOffchainSecrets` to be even more explicit.

README.md
KuphJr: Because the spacing changes.  The README columns adjust to fit the longest line of text, hence the change.

pinebit (approved):

PR #73, opened 2023-02-07
Add OpenZeppelin Upgradable

justinkaseman:
- [x] Add OpenZeppelin Upgradable
- [x] Update deployment command
- [x] Update local deployment within simulate command 
- [x] Add upgrade contract command
- [x] Sepolia deployment
- [x] Mumbai deployment
- [x] Remove Goerli support


tasks/Functions-admin/deployFullOracle.js
```diff
@@ -22,33 +22,40 @@ task(
...
-  const registry = overrides
-    ? await registryFactory.deploy(linkTokenAddress, linkEthFeedAddress, oracle.address, overrides)
-    : await registryFactory.deploy(linkTokenAddress, linkEthFeedAddress, oracle.address)
-  console.log(`Waiting for transaction ${registry.deployTransaction.hash} to be confirmed...`)
-  await registry.deployTransaction.wait(1)
-  console.log(`FunctionsBillingRegistry deployed to ${registry.address} on ${network.name}`)
+  const registryProxy = overrides
+    ? await upgrades.deployProxy(registryFactory, [linkTokenAddress, linkEthFeedAddress, oracleProxy.address], {
+        kind: "transparent",
+        constructorArgs: overrides,
```
KuphJr, line 50: `constructorArgs` seems like a weird name, but I assume this is just forwarding the overrides to the underlying `ethers` call.
Just double-checking this is the correct way to specify overrides.

justinkaseman: I know what you mean, It'll get tested it when I do the Goerli deployment. If it isn't `constructorArgs` then I'm not sure that Hardhat Upgrades exposes a way to override deployment settings.

tasks/Functions-admin/deployFullOracle.js
```diff
@@ -70,62 +77,48 @@ task(
   console.log("Registry configuration set")
 
   // Set up Functions Oracle
```
KuphJr, line 79: Do we only need to call `acceptOwnership` when the proxy contract is deployed and not when the contract is upgraded?

Also, I assume that we are deploying the proxy contracts manually first, then entering the proxy addresses into `network-config.js` before running `functions-deploy-oracle`?

tasks/Functions-admin/upgrade.js
```diff
@@ -0,0 +1,69 @@
...
+      overrides = {
+        // be careful, this may drain your balance quickly
+        maxPriorityFeePerGas: ethers.utils.parseUnits("50", "gwei"),
+        maxFeePerGas: ethers.utils.parseUnits("50", "gwei"),
+      }
+    }
+
+    let proxyAddress, newImplementationFactory, contractPath
+    const { type } = taskArgs
+    if (type === "registry" || type === "Registry" || type === "REGISTRY") {
```
KuphJr, line 25: The `if` and `else if` appear to have identical logic

tasks/Functions-admin/deployFullOracle.js
KuphJr: Thanks for double-checking!

tasks/Functions-admin/deployFullOracle.js
KuphJr: I am still not completely sure I understand the deployment logic, but I will spend more time reviewing & ask relevant questions once this PR moves from `draft` to `open`.

It might be worth a 15 minute sync to explain too.  Whatever is easier.

tasks/Functions-admin/upgrade.js
justinkaseman: Oops, good catch

tasks/Functions-admin/deployFullOracle.js
justinkaseman: Let's sync because deployment timing is a little weird.

justinkaseman: > Do we only need to call acceptOwnership when the proxy contract is deployed and not when the contract is upgraded?

Yes. The hardhat upgrades plugin handles all of the ownership for us.

> Also, I assume that we are deploying the proxy contracts manually first, then entering the proxy addresses into network-config.js before running functions-deploy-oracle?

No, `functions-deploy-oracle` deploys the proxy admin, proxy, and imlementation - then it spits back out the addresses for us to updates `network-config.js`.

KuphJr:
We still have to get rid of the Goerli gas overrides (and any other mention of Goerli) before this is merged, right?

tasks/Functions-admin/deployFullOracle.js
justinkaseman: Removed these, they indeed do not work.

justinkaseman:
TODO: re-do mumbai deployment

tasks/Functions-admin/setOcrConfig.js
```diff
@@ -7,38 +7,28 @@ task("functions-set-ocr-config", "Sets the OCR config using values from a given
...
+      )
+      throw new Error(error)
     }
 
-    const oracleFactory = await ethers.getContractFactory("FunctionsOracle")
-    const oracle = oracleFactory.attach(networkConfig[network.name]["functionsOracle"])
+    const oracleFactory = await ethers.getContractFactory("contracts/dev/functions/FunctionsOracle.sol:FunctionsOracle")
+    const oracle = oracleFactory.attach(networkConfig[network.name]["functionsOracleProxy"])
 
-    const ocrConfig = require("../../" + taskArgs.configfile)
```
bolekk, line 22: Heads up - we don't want to delete that.

justinkaseman: Oops, bad merge conflict resolution. Good catch.

justinkaseman:
Status: Awaiting NOPs to update to latest Mumbai deployment

contracts/dev/functions/migrations/FunctionsOracleMigration.sol
```diff
@@ -0,0 +1,272 @@
+// SPDX-License-Identifier: MIT
```
bolekk, line 1: I forgot why we need the migrations dir. Can't we just upgrade the contracts and merge into main afterwards?

network-config.js
```diff
@@ -27,33 +27,22 @@ const networkConfig = {
...
   },
   mumbai: {
     linkToken: "0x326C977E6efc84E512bB9C30f76E30c160eD06FB",
     linkEthPriceFeed: "0x12162c3E810393dEC01362aBf156D7ecf6159528",
-    functionsOracle: "0x6199175d137B791B7AB06C3452aa6acc3519b254",
-    functionsOracleFactory: "0xCBEAb3ECc3Af26bE11E25767bc6DDA5f0B129BEA",
-    functionsOracleRegistry: "0xB12044Ba63F66191E53b0Cd8C10095080b4c8434",
+    functionsOracleProxy: "0xeA6721aC65BCeD841B8ec3fc5fEdeA6141a0aDE4",
+    functionsBillingRegistryProxy: "0xEe9Bf52E5Ea228404bB54BCFbbDa8c21131b9039",
     functionsPublicKey:
```
bolekk, line 38: Update both public keys to:

a30264e813edc9927f73e036b7885ee25445b836979cb00ef112bc644bd16de2db866fa74648438b34f52bb196ffa386992e94e0a3dc6913cee52e2e98f1619c

tasks/Functions-admin/upgrade.js
```diff
@@ -0,0 +1,60 @@
...
+    console.log("\n__Recompiling Contracts__")
+    await run("compile")
+
+    if (network.name === "hardhat") {
+      throw Error("This command cannot be used on a local development chain.  Specify a valid network.")
+    }
+
+    let proxyAddress, newImplementationFactory, contractPath
+    const { type } = taskArgs
+    if (type === "registry" || type === "Registry" || type === "REGISTRY") {
```
bolekk, line 16: How about 
`if (type.toLowerCase() === "registry")`

bolekk (changes requested):
One last pass.

.gitignore
```diff
@@ -103,4 +103,7 @@ gas-report.txt
 contracts/test/fuzzing/crytic-export
 
 Functions-request.json
-offchain-secrets.json
\ No newline at end of file
+offchain-secrets.json
+
+# OpenZeppelin Upgrades local network artifacts
+.openzeppelin/unknown-31337.json
```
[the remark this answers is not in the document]
justinkaseman: The artifacts are what the hardhat plugin uses to determine if an upgrade has changes that break the proxy pattern.

tasks/Functions-admin/upgrade.js
justinkaseman: That does sound cleaner. Changing.

contracts/dev/functions/FunctionsClient.sol
```diff
@@ -57,10 +57,9 @@ abstract contract FunctionsClient is FunctionsClientInterface {
   function sendRequest(
     Functions.Request memory req,
     uint64 subscriptionId,
-    uint32 gasLimit,
```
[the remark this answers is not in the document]
bolekk: You know that I don't like merging different changes together either. In this case Morgan's change was already merged to core and not migrated here. Plus this PR already combines a few other pending changes to contracts. I want to minimize time during which contracts differ between the two repos.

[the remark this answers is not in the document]
justinkaseman: Right - I forgot about that it's already in core :man_facepalming:

bolekk (approved):
LGTM!

PR #81, opened 2023-02-10
Update buildRequestJSON.js

zeuslawyer:
It appears that this task is designed for local machine development, so the network guard seems wrong.


KuphJr:
This command needs to connect to the blockchain to pull the secrets encryption public keys.

PR #83, opened 2023-02-10
Increase request tx gas limit

KuphJr:
One of dev rel demos hit the 500k gas limit for a request due to the size of the JavaScript source of ~150 lines.  This should be increased to prevent others from encountering this issue.


tasks/Functions-client/request.js
```diff
@@ -18,7 +18,7 @@ task("functions-request", "Initiates a request from an Functions client contract
   .setAction(async (taskArgs, hre) => {
     // A manual gas limit is required as the gas limit estimated by Ethers is not always accurate
     const overrides = {
-      gasLimit: 500000,
+      gasLimit: 1500000,
```
pinebit, line 21: I remember we had a task argument that could override this. Can't we use a task argument and fall back to 50K?

pinebit (changes requested):

tasks/Functions-client/request.js
KuphJr: The gaslimit argument is for the amount of gas used for the callback transaction, not the transaction which initiates the request.

KuphJr: However, I can add an optional `requestgas` task arg.

pinebit: You are right. Long term I believe we should always take hardhat.config.js value from the corresponding network{}, if specified. gasLimit is supposed to be very different for sepolia and mumbia and this will cause confusion for many people switching network.

pinebit (approved):
Approving to speeed up on development, but I am not very happy with the second argument. Very easy to misinterpret. As stated in my other comment, I prefer all overrides to come from network config (hardhat config js).

PR #86, opened 2023-02-12
added option to disable simulation before request

KuphJr:


KuphJr:
> When would someone do this?

@zeuslawyer mentioned it would be a nice addition.  I know he was testing the number of queries made to the API he was using at the time, but he may be able to lend insight to other reasons a dev might want this.

zeuslawyer:
@KuphJr @justinkaseman  so far its only to help reduce the number of API hits and on the basis that a dev may simulate a few times and then *may* want to skip the sim when they're only wanting to see how it runs on the actual testnet.

justinkaseman (approved):

PR #93, opened 2023-02-16
readme improvements & removed mention of unused REPORT_GAS env var

KuphJr:


contract-diff.sh
```diff
@@ -1,23 +0,0 @@
-#!/bin/zsh
```
bolekk, line 1: Let's keep this. It's just a sanity check between two public repos. You can actually add that comment on top of the file.

bolekk (approved):

PR #100, opened 2023-02-22
Feat/request timeout

KuphJr:


tasks/Functions-billing/timeoutRequest.js
```diff
@@ -0,0 +1,24 @@
+const { networkConfig } = require("../../network-config")
+
+task("functions-timeout-requests", "Times out expired requests")
+  .addParam("requestids", "1 or more request IDs to timeout")
```
justinkaseman, line 4: What format are these coming in as? Comma separated string?
We should be more explicit either through the description or actually enforcing types

tasks/Functions-billing/timeoutRequest.js
```diff
@@ -0,0 +1,24 @@
...
+
+    console.log(
+      `Timing out requests ${taskArgs.requestids} on ${taskArgs.network}`
+    )
+
+    const RegistryFactory = await ethers.getContractFactory("FunctionsBillingRegistry")
+    const registry = await RegistryFactory.attach(networkConfig[network.name]["functionsBillingRegistryProxy"])
+
+    await registry.timeoutRequests(taskArgs.requestids)
+
```
justinkaseman, line 20: Best practice to wait a block or a couple blocks to confirm as well

justinkaseman (commented):
Some nits

tasks/Functions-billing/timeoutRequest.js
KuphJr: thanks.  I couldn't actually test this as it would require somehow forcing a request to fail, so I appreciate the diligent review.

justinkaseman: You should still be able to call the function, even if the requests aren't timed out

tasks/Functions-billing/timeoutRequest.js
```diff
@@ -0,0 +1,24 @@
...
+    console.log(
+      `Requests ${taskArgs.requestids} are no longer pending or have been timed out`
```
justinkaseman, line 22: A more foolproof way to do this might be to check for an event to be emitted, similar to how the `request` command does.

Looks like timed out requests will `emit RequestTimedOut(requestId);`

justinkaseman (approved):
LGTM

PR #115, opened 2023-03-27
Created required functions

KuphJr:
This PR adds the necessary functions to upload off-chain encrypted secrets to GitHub

### DIFF
--- a/tasks/utils/github.js
+++ b/tasks/utils/github.js
@@ -1,0 +1,70 @@
+const axios = require("axios")
+
+export const createGist = async (githubApiToken, encryptedOffchainSecrets) => {
+  const content = JSON.stringify(encryptedOffchainSecrets)
+
+  const headers = {
+    Authorization: `token ${githubApiToken}`,
+  }
+
+  // construct the API endpoint for creating a Gist
+  const url = "https://api.github.com/gists"
+  const body = {
+    public: true,
+    files: {
+      [`encrypted-functions-request-data-${new Date().now()}.json`]: {
+        content,
+      },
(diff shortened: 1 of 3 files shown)


tasks/utils/github.js
```diff
@@ -0,0 +1,70 @@
...
+  const content = JSON.stringify(encryptedOffchainSecrets)
+
+  const headers = {
+    Authorization: `token ${githubApiToken}`,
+  }
+
+  // construct the API endpoint for creating a Gist
+  const url = "https://api.github.com/gists"
+  const body = {
+    public: true,
```
justinkaseman, line 13:
```suggestion
    public: false,
```
We want private gists

tasks/utils/github.js
```diff
@@ -0,0 +1,70 @@
...
+      },
+    },
+  }
+
+  try {
+    const response = await axios.post(url, body, { headers })
+    const gistUrl = response.data.html_url
+    return gistUrl
+  } catch (error) {
+    console.error("Failed to create Gist", error)
```
justinkaseman, line 26: We need to throw again here right? Rather than swallow it
Otherwise output of this function will be undefined

tasks/utils/github.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+}
+
+export const checkTokenGistScope = async (githubApiToken) => {
```
justinkaseman, line 30: :heart:

justinkaseman (changes requested):
Two quick changes, otherwise looks good

PR #116, opened 2023-03-27
Chore/update functions sandbox library

KuphJr:
This PR does not affect the functionality of the tooling repo, but preps for the removal of on-chain secrets in favor of exclusively supporting off-chain secrets by updating the FunctionsSandboxLibrary

### DIFF
--- a/FunctionsSandboxLibrary/buildRequest.js
+++ b/FunctionsSandboxLibrary/buildRequest.js
@@ -1,17 +1,17 @@
-"use strict"
-Object.defineProperty(exports, "__esModule", { value: true })
-exports.buildRequest = void 0
-const getRequestConfig_1 = require("./getRequestConfig")
-const encryptSecrets_1 = require("./encryptSecrets")
+"use strict";
+Object.defineProperty(exports, "__esModule", { value: true });
+exports.buildRequest = void 0;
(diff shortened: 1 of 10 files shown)


FunctionsSandboxLibrary/buildRequest.js
```diff
@@ -6,38 +6,8 @@ const encryptSecrets_1 = require("./encryptSecrets")
 const buildRequest = async (unvalidatedConfig) => {
   const config = (0, getRequestConfig_1.getRequestConfig)(unvalidatedConfig)
   const request = { source: config.source }
-  if (
```
justinkaseman, line 9: Just to confirm, this gets merged then the sandbox will no longer handle inline secrets right?
So it will be a breaking change to the `request` command trying to use inline secrets

FunctionsSandboxLibrary/getRequestConfig.js
```diff
@@ -19,65 +19,61 @@ var CodeLanguage
 const getRequestConfig = (unvalidatedConfig) => {
   const config = unvalidatedConfig
   if (config.codeLocation !== Location_.Inline) {
-    throw Error(`codeLocation is not correctly specified in config file`)
```
justinkaseman, line 22: What is the relevance of the word "file" here?

PR #118, opened 2023-03-27
Implemented Github remote secrets for `functions-request` command

KuphJr:
Completed:
- Verify token scope & warn user if it is too broad
- Upload Gist containing encrypted secrets
- Use Gist URL to provide secrets in an on-chain request
- Delete Gist once request is complete
- Implemented automatic Gist uploading for Automation contract deployment
- Updated task commands for new contracts without inline secrets
- Update documentation to provide information about Gist secret support & required GitHub API token


tasks/Functions-client/request.js
```diff
@@ -100,10 +103,11 @@ task("functions-request", "Initiates a request from a Functions client contract"
     }
 
     const transactionEstimateGas = await clientContract.estimateGas.executeRequest(
-      request.source,
-      request.secrets ?? [],
-      requestConfig.secretsLocation,
-      request.args ?? [],
+      requestConfig.source,
+      requestConfig.secrets && Object.keys(requestConfig.secrets).length > 0 ? simulatedSecretsURLBytes : [],
+      // TODO: update contracts so secretsLocation is no longer required
```
justinkaseman, line 108: Can change now

tasks/Functions-client/request.js
```diff
@@ -222,7 +235,7 @@ task("functions-request", "Initiates a request from a Functions client contract"
       const requestTx = await clientContract.executeRequest(
         request.source,
         request.secrets ?? [],
-        requestConfig.secretsLocation,
+        1,
```
justinkaseman, line 238: Can delete now

justinkaseman (approved):
Overall LGTM, only thing I can see to change is the clientContract.executeRequest calls

I'll update automatedConsumer in the morning using the `RequestStore`

PR #119, opened 2023-03-27
Utility for storing artifacts about a Functions request

justinkaseman:
Data structure will become more solid once added to the `request` command

### Caveat
Likely will need added handling of local networks to be able to use in a testing environment


KuphJr (approved):
Base class LGTM!  I might have some more helpful input once we start integrating this into the existing tasks.

PR #120, opened 2023-03-28
Update to latest contracts

justinkaseman:
Update to latest @chainlink/contracts

### Changelog
    - Amend transmitter double DON fee portion payout
    - Return FulfillResult as output from `fulfillAndBill`
    - Optimize AuthorizedOriginReceiverUpgradable
    - Functions library uses solidty-cborutils CBOR v2.0.0 and ENS Buffer v0.1.0(#8485)
    - Remove support for inline secrets in Functions requests


contracts/dev/functions/FunctionsBillingRegistry.sol
```diff
@@ -291,7 +290,7 @@ contract FunctionsBillingRegistry is
     FunctionsBillingRegistryInterface.RequestBilling memory /* billing */
   ) public pure override returns (uint96) {
     // NOTE: Optionally, compute additional fee here
-    return 200_000_000_000_000_000; // 0.2 LINK
+    return 0;
```
KuphJr, line 293: Does this need to be set back to 0.2 LINK?

KuphJr (commented):
Just had a question about the fee. Otherwise LGTM

contracts/dev/functions/FunctionsBillingRegistry.sol
justinkaseman: Yep

justinkaseman:
```suggestion
      return 200_000_000_000_000_000; // 0.2 LINK
```

KuphJr (approved):

PR #121, opened 2023-03-28
Use request artifacts

justinkaseman:
Use `RequestStore` to store information about sent requests

### DIFF
--- a/tasks/Functions-client/request.js
+++ b/tasks/Functions-client/request.js
@@ -4,6 +4,7 @@ const { VERIFICATION_BLOCK_CONFIRMATIONS, networkConfig } = require("../../netwo
 const utils = require("../utils")
 const chalk = require("chalk")
 const { deleteGist } = require("../utils/github")
+const { RequestStore } = require("../utils/artifact")
 
 task("functions-request", "Initiates a request from a Functions client contract")
   .addParam("contract", "Address of the client contract to call")
@@ -79,9 +80,9 @@ task("functions-request", "Initiates a request from a Functions client contract"
     const { lastBaseFeePerGas, maxPriorityFeePerGas } = await hre.ethers.provider.getFeeData()
     const estimatedCostJuels = await clientContract.estimateCost(
       [
-        0, // Inline
-        1, // Remote
-        0, // JavaScript
+        requestConfig.codeLocation,
+        1, // SecretsLocation: Remote
+        requestConfig.codeLanguage,
         requestConfig.source,
         requestConfig.secrets && Object.keys(requestConfig.secrets).length > 0 ? simulatedSecretsURLBytes : [],
         requestConfig.args ?? [],
@@ -127,6 +128,8 @@ task("functions-request", "Initiates a request from a Functions client contract"
     const request = await generateRequest(requestConfig, taskArgs)
     doGistCleanup = doGistCleanup && request.secrets
 
+    const store = RequestStore(hre.network.config.chainId, network.name)
+
     console.log("\n")
     const spinner = utils.spin({
       text: `Submitting transaction for FunctionsConsumer contract ${contractAddr} on network ${network.name}`,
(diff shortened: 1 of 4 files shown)


tasks/Functions-client/request.js
```diff
@@ -136,21 +139,37 @@ task("functions-request", "Initiates a request from a Functions client contract"
     await new Promise(async (resolve, reject) => {
       let requestId
 
+      const cleanup = async () => {
```
justinkaseman, line 142: Made this more consistent

tasks/utils/artifact.js
```diff
@@ -4,51 +4,54 @@ const fs = require("node:fs/promises")
 const currentVersion = "0"
 
 // export interface RequestData {
```
justinkaseman, line 6: Could also hold billing data - but keeping it leaner for now. If we want it later we can add to this structure and use the migration functions.

tasks/utils/artifact.js
```diff
@@ -4,51 +4,54 @@ const fs = require("node:fs/promises")
...
+  constructor(
+    chainId /*: number */,
+    chainName /*: string */,
+    directory = DEFAULT_DIRECTORY,
+    subdirectory = DEFAULT_SUBDIRECTORY
+  ) {
     this.chainId = chainId
-    this.path = path.join(directory, subdirectory, String(chainId))
+    this.chainName = chainName
+    const network = `${chainId}-${chainName}`
```
justinkaseman, line 53: Makes the folder a little nicer to read

tasks/utils/artifact.js
```diff
@@ -4,51 +4,54 @@ const fs = require("node:fs/promises")
...
-//  // args (string only array) can be accessed within the source code with `args[index]` (ie: args[0]).
-//   args: ["1", "bitcoin", "btc-bitcoin", "btc", "1000000", "450"],
-//  // expected type of the returned value
-//   expectedReturnType: ReturnType.uint256,
-//  // Default offchain secrets object used by the `functions-build-offchain-secrets` command
-//   globalOffchainSecrets: {},
-//  // Per-node offchain secrets objects used by the `functions-build-offchain-secrets` command
-//   perNodeOffchainSecrets: [],
+//   type: 'consumer' | 'automatedConsumer'
+//   id: string;
```
justinkaseman, line 8: Notice that requests can either be:
type: 'consumer', which use request id as their id
or 
type: 'automatedConsumer', which use the AutomatedConsumer contract address as their id

tasks/utils/artifact.js
justinkaseman: @KuphJr Anything else we immediately want in here? If so, why?

tasks/utils/artifact.js
KuphJr: What is the best way to make this clear to the user?  Could we have `requestId` and `automationContractAddress` instead? Maybe we could even separate Automation requests into a separate file from regular requests?

I think it may be confusing that `id` has different meanings in different contexts.

tasks/utils/artifact.js
KuphJr: No, I think this is sufficient for now.

tasks/utils/artifact.js
```diff
@@ -58,38 +61,38 @@ class RequestStore {
 
   async create(data /*: RequestData*/) /*: Promise<void> */ {
     validateDataVersion0(data)
-    if (await this.exists(data.requestId)) {
-      throw new Error(`Request ${data.requestId} already exists on chain ${this.chainId}`)
+    if (await this.exists(data.id)) {
+      throw new Error(`Request ${data.id} already exists on chain ${this.chainId}`)
```
KuphJr, line 65: Do we want to throw an error here or log a warning?  Also, since the user could call `setAutoRequest` multiple times for the same contract to update the request, will this cause an unintended error?

KuphJr (commented):
Some notes

KuphJr (commented):
Some notes

tasks/Functions-client/request.js
```diff
@@ -173,8 +192,8 @@ task("functions-request", "Initiates a request from a Functions client contract"
         }
         ocrResponseEventReceived = true
         if (billingEndEventReceived) {
-          spinner.stop()
-          return resolve()
+          await store.update(requestId, { status: "complete", result })
```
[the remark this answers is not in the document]
justinkaseman: Instead moving this up a line because setting the result and completion is tied to `OCRResponse`

tasks/utils/artifact.js
justinkaseman: I agree. Broke them apart into two different folders.

tasks/utils/artifact.js
justinkaseman: Kept it as throwing and added `.upsert` method for automation requests

KuphJr (approved):
Will need a rebase & a quick test depending on if the larger `feat/remove-inline-secrets` PR is merged first, but otherwise it LGTM!

PR #129, opened 2023-03-31
Add 'functions-clear-gists' command

justinkaseman:


tasks/Functions-client/clearGists.js
```diff
@@ -0,0 +1,36 @@
...
+      return console.log(`Automated Consumer ${taskArgs.contract} not found.`)
+    }
+
+    if (!gist.activeManagedSecretsURLs || gist.secretsURLs.length < 1) {
+      return console.log(`Automated Consumer ${taskArgs.contract} does not have active secret Gists.`)
+    }
+
+    await Promise.all(
+      gist.secretsURLs.map(async (url) => {
+        return await deleteGist(process.env["GITHUB_API_TOKEN"], url.slice(0, -4))
```
bolekk: Why url.slice(0, -4) ? Can we put it in a var with a descriptive name?

tasks/utils/artifact.js
```diff
@@ -98,18 +92,21 @@ class RequestStore {
...
 
-  async upsert(id /*: string*/, data /*: Fragment<RequestData>*/) /*: Promise<void> */ {
+  async upsert(id /*: string*/, data /*: Fragment<RequestData>*/) /*: Promise<boolean> */ {
+    let created = false
     let previousData = {}
     let newData = data
     try {
       previousData = await this.read(id)
     } catch {
+      created = true
```
bolekk, line 102: Do we always expect an exception here?

tasks/utils/artifact.js
```diff
@@ -83,7 +78,6 @@ class RequestStore {
       const data = JSON.parse(await this.readFile(id))
       return validateRequestArtifactVersion(data)
     } catch (e) {
-      console.error(`Could not find artifact with the ID of ${id}`)
```
bolekk, line 86: Do we need to catch if we always re-throw?

tasks/utils/artifact.js
justinkaseman: Yes, `.read` will throw if the file is not found.

bolekk (approved):

PR #134, opened 2023-04-05
Update to support latest version of `env-enc`

KuphJr:
**Done**
- Updated README for latest version of `env-enc`
- Add `env-enc` as a dependency once it is published to NPM


package.json
```diff
@@ -48,6 +48,7 @@
     "typechain": "^8.1.0"
   },
   "dependencies": {
+    "@chainlink/env-enc": "^1.0.1",
```
justinkaseman, line 51:
```suggestion
    "env-enc": "^1.0.1",
```
(for when published)

PR #139, opened 2023-04-14
Fixed functions-simulate command

pinebit:


.solhint.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "solhint:recommended",
   "rules": {
-    "compiler-version": ["error", "^0.8.0"],
+    "compiler-version": ["warn", "^0.8.0"],
```
pinebit, line 4: @justinkaseman I could not fix this. Please help me getting this back to `error`.

KuphJr, line 4: Are we able to disable this entirely?  I know some novice users can become confused by these version warnings and I do not believe they are very important for the vast majority of users.

KuphJr:
I would like to see if we can avoid the version warnings entirely, but I do not believe it is a blocker for merging this PR, given there are many Solidity compiler warnings already that users should ignore.

I will let Justin give final approval for the PR before merging, just in kase 😉 he knows the solution.

PR #142, opened 2023-04-26
Networks refactoring + added fuji

pinebit:


README.md
```diff
@@ -112,7 +125,7 @@ The `--path` flag has no effect on the `npx env-enc set-pw` command as the passw
 The Functions and Functions subscription management commands commands can be executed in the following format:
 `npx hardhat command_here --parameter1 parameter_1_value_here --parameter2 parameter_2_value_here`
 
-Example: `npx hardhat functions-read --network mumbai --contract 0x787Fe00416140b37B026f3605c6C72d096110Bb8`
+Example: `npx hardhat functions-read --network polygonMumbai --contract 0x787Fe00416140b37B026f3605c6C72d096110Bb8`
```
KuphJr, line 128: I am a little concerned about causing confusion due to changing the network name.  Do you think the verbose name is worth this potential confusion?  If so, you will need to sync with @aelmanaa so he can update the docs accordingly.

pinebit: I have already created a ticket to update docs: https://smartcontract-it.atlassian.net/browse/FUN-368
Besides this "renaming", we will need to include the new network details in the public docs.

KuphJr: Great! The rest looks good to me other than the missing block explorer, however if you prefer you can address that separately.  I am going to approve this now.

KuphJr (approved):

networks.js
```diff
@@ -0,0 +1,61 @@
...
+    functionsBillingRegistryProxy: "0xEe9Bf52E5Ea228404bB54BCFbbDa8c21131b9039",
+    functionsPublicKey: SHARED_DON_PUBLIC_KEY,
+  },
+  avalancheFuji: {
+    url: process.env.AVALANCHE_FUJI_RPC_URL || "UNSET",
+    accounts: PRIVATE_KEY !== undefined ? [PRIVATE_KEY] : [],
+    chainId: 43113,
+    confirmations: 2 * DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS,
+    linkToken: "0x0b9d5D9136855f6FEc3c0993feE6E9CE8a297846",
+    linkEthPriceFeed: "0xf4060f80f295b34e0C2471461ba43745Aeb186d6",
```
justinkaseman, line 51: This should be AVAX/LINK, not the ETH/LINK feed.

Also applies to the actual contract's configuration

pinebit: Good point! I will address this in the next PR.

PR #148, opened 2023-05-03
Mitigate gas spikes & delete Gist upon transaction failure

KuphJr:


networks.js
```diff
@@ -22,6 +22,7 @@ if (!isTestEnvironment && !PRIVATE_KEY) {
 const networks = {
   ethereumSepolia: {
     url: process.env.ETHEREUM_SEPOLIA_RPC_URL || "UNSET",
+    gasPrice: undefined,
```
bolekk, line 25: Where is it used in the code?

pinebit: In hardhat.config.js: `...networks,`
this is a standard hardhat parameter that is a part of network configuration.

pinebit (approved):

networks.js
KuphJr: By default, leaving these values unspecified uses standard gas price estimation.  However, I wanted to show these fields so it is obvious to users what they have to change after reading the README.

Issue #158, opened 2023-05-12
Simulation being executed ONLY AFTER user confirmed desire to continue with transaction


mnedelko:
I used a mix of docs, readme and the following VDO (https://youtu.be/nKRZ1R2BC4Q?t=587) to familiarize myself with Chainlink functions. 

While doing so I came across the following behavior, which appears to deviate from the behavior outlined in the docs/tutorial videos. I hope I am not wrong about this ;) :

EXPECTED BEHAVIOR:
1. Executed `npx hardhat functions-request --network polygonMumbai --contract [contractid] --subid [subid]`
2. Code contained in the function will be simulated and run locally
3. CLI will provide a simulated result
4. CLI will THEN estimate fees and asks user to confirm that they are willing to proceed

OBSERVED BEHAVIOR (see screenshot below):
1. Executed `npx hardhat functions-request --network polygonMumbai --contract [contractid] --subid [subid]`
2. CLI estimated fees and asks user to confirm that they are willing to proceed
3. Only AFTER the user responded with YES will the function be simulated and run locally
4. CLI will provide a simulated result

![functions_observed](https://github.com/smartcontractkit/functions-hardhat-starter-kit/assets/6106392/faeb1ec9-fc7c-4a9a-85ac-ec22a88d30db)

Other than that, its a great example, and a wonderful 'getting started' experience.

mnedelko:
Looks like the following line in `~/tasks/Functions-client/request.js` got pushed down in the order of execution during the following commit: https://github.com/smartcontractkit/functions-hardhat-starter-kit/commit/0bd72821c5343a1266860382602efdf40712ca9c 

Line in question:
`const request = await generateRequest(requestConfig, taskArgs)`

Will try to submit a PR to propose a fix.

KuphJr:
Thank you for raising this issue!
I think we need to make a few changes in order for this logic to work for both single Functions requests and Automation Functions requests.

I have filed an internal ticket to resolve.

KuphJr:
Resolved in `develop`
Thank you for raising!

PR #159, opened 2023-05-12
#158 Proposed Fix: Changed the order of execution and moved up generateRequest call

mnedelko:
Changed the order of execution and moved up generateRequest call to ensure request.js simulates results before estimating transaction costs and before providing the user with an opportunity to confirm they are willing to proceed


KuphJr:
Thanks for the PR! This will be addressed shortly in a separate PR with some other improvements, so I am closing this one.

PR #161, opened 2023-06-03
ability to deploy javascript code on chain to a contract

dbeal-eth:
currently, it is required to include the code to be executed inline when requesting functions execution. this is great, as it means that the code to be executed has to be on-chain to be trusted. however, as contracts and their corresponding offchain code gets more complicated, it becomes necesary to store the code in a separate contract in order to stay under the EVM size limit. Fortunately, it is actually possible to push a non-functional contract embedde with the blob of the code directly on-chain. Its a reaonsably efficient way to do so as a execution node (for ethereum or otherwise) is able to store contract source data contiguously in storage, allowing for efficient load by `extcode` or whatever.

This PR proposed addition of a function which will allow for the specified javascript functions file to be uploaded on-chain as a contract. the bytecode of the contract will effectively just be the contents of the javascript file, and these can be loaded by a Functions Consumer for inline execution by a function similar to the following:

```

    /**
     * loads the bytes deployed to the specified address
     * used to get the inline execution code for chainlink
     */
    function _codeAt(address _addr) public view returns (bytes o_code) {
        assembly {
            // retrieve the size of the code, this needs assembly
            let size := extcodesize(_addr)
            // allocate output byte array - this could also be done without assembly
            // by using o_code = new bytes(size)
            o_code := mload(0x40)
            // new "memory end" including padding
            mstore(0x40, add(o_code, and(add(add(size, 0x20), 0x1f), not(0x1f))))
            // store length in memory
            mstore(o_code, size)
            // actually retrieve the code, this needs assembly
            extcodecopy(_addr, add(o_code, 0x20), 0, size)
        }
    }
```

(alternatively, functions could instead encode the address of the code into the CBOR and save the consumer 10k gas or so per call, but one step at a time :laugh:)

this same pattern could be applied to secrets as well.


zeuslawyer:
@dbeal-eth - interesting idea and thanks for taking the time to share these thoughts!   Have you done this locally yourself using that assembly code you suggested?  Have you encountered a situation where your contract was over the EVM size limit? Would love to know more about that experience + usecase.

dbeal-eth:
> @dbeal-eth - interesting idea and thanks for taking the time to share these thoughts!   Have you done this locally yourself using that assembly code you suggested?  Have you encountered a situation where your contract was over the EVM size limit? Would love to know more about that experience + usecase.

Yes, this feature was developed because it was no longer possible to store the functions code inline due to contract size limit, and storing the code in the function storage is inefficient gas-wise. I have used the assembly code both locally and on-chain with sepolia and Mumbai.

dbeal-eth:
my apologies, I should have submitted the contracts that I deployed on-chain as examples.

On sepolia, I deployed https://sepolia.etherscan.io/address/0x22B6f14F822f49AE5Df6566138b3359358444f22
On mumbai, I deployed https://mumbai.polygonscan.com/address/0xB28E22e615B06EAd0272136C2C6dD6adB2aaD34D

if you just take the hex code of either of these contracts and decode them, you will see that its not a real solidity contract/EVM bytecode, but actually just hex encoded javascript file.

KuphJr:
Thank you! In order to keep this repo focused on basic demos, this PR will be closed, but please feel free to fork and share you example with others!

PR #167, opened 2023-06-26
Added ref links to get private key and RPC URL

darrensapalo:
- Added additional link to explain the concept of private keys (and how to acquire them).
- Added additional link on where you can possibly acquire the RPC URLs by visiting the Chainlist website.


KuphJr:
Thanks for the contribution!  However, we don't want to link to too many 3rd party sites in the docs if we can avoid it, so I am closing for now.  These clarifications should eventually be addressed in our [official Functions docs](https://docs.chain.link/chainlink-functions).

PR #174, opened 2023-08-25
Refactor to use NPM tooling for funcs - Sub Manager

zeuslawyer:
Subscription Management tasks updated.

1. Staging only support Mumbai so updated networks.js
2. updated create.js with subscription funding prompt and using NPM package.
3. updated fund.js on the assumption that toolkit will now return sub balance as a BigNum string, not number.
4. updated add.js to add consumers.
5. updated info.js to implement get sub info.
6. cancel.js done.
7. sub transfer and acceptance of transfer done.
8. remove consumer done.

PR #175, opened 2023-08-26
Refactor to use NPM tooling for funcs. Subscription Management tasks …

zeuslawyer:
…updated.

1. Staging only support Mumbai so updated networks.js
2. updated create.js with subscription funding prompt and using NPM package.
3. updated fund.js on the assumption that toolkit will now return sub balance as a BigNum string, not number.
4. updated add.js to add consumers.
5. updated info.js to implement get sub info.
6. cancel.js done.
7. sub transfer and acceptance of transfer done.
8. remove consumer done.

NOTE:  original PR: https://github.com/smartcontractkit/functions-hardhat-starter-kit/pull/174
This PR addresses comments from that PR.


tasks/Functions-billing/fund.js
```diff
@@ -8,60 +11,32 @@ task("functions-sub-fund", "Funds a billing subscription for Functions consumer
...
-    // Ensure sufficient balance
-    const balance = await linkToken.balanceOf(signer.address)
-    if (juelsAmount.gt(balance)) {
-      throw Error(
-        `Insufficient LINK balance. Trying to fund subscription with ${ethers.utils.formatEther(
-          juelsAmount
-        )} LINK, but wallet only has ${ethers.utils.formatEther(balance)}.`
-      )
-    }
+    console.log(`\nFunding subscription '${subId}' with '${linkAmount}' LINK...`)
```
KuphJr, line 31: convert linkAmount -> juelsAmount

tasks/Functions-billing/fund.js
```diff
@@ -8,60 +11,32 @@ task("functions-sub-fund", "Funds a billing subscription for Functions consumer
...
 
-    // Fund the subscription with LINK
-    const fundTx = await linkToken.transferAndCall(
-      networks[network.name]["functionsBillingRegistryProxy"],
-      juelsAmount,
-      ethers.utils.defaultAbiCoder.encode(["uint64"], [subscriptionId])
-    )
+    const fundTxReceipt = await sm.fundSubscription({ juelsAmount, subId, txOptions })
+    console.log(`\nSubscription '${subId}' funded with '${linkAmount}' LINK in Tx: ''${fundTxReceipt.transactionHash}'`)
```
KuphJr, line 34: Let's create a util function that gets the output from `getSubscriptionInfo` and converts the balance amounts from Juels to LINK, returning the SubInfo object with the converted values.  Then, we can use that util function anywhere we use `getSubscriptionInfo` before displaying to the user.

KuphJr:
Once we address the LINK->Juels conversion we should be ready to merge.

.gitignore
```diff
@@ -11,7 +11,7 @@ coverage.json
 typechain
 
 # don't push the environment vars!
-.env
+*.env
 
 # These environment variables are encrypted, but should not be pushed unless a secure password is used
 .env.enc
```
KuphJr, line 17: You can remove this since you match above with *.env

zeuslawyer: @KuphJr i added the * because currently its a hard match on .env only.  My notes.env file was getting staged

tasks/Functions-billing/fund.js
zeuslawyer: as per slack chat, if we wrap sm.fundSubscription in a helper function we lose educational value on the NPM package as the end user may not dig into the abstraction.  This way we show the sm to the end user directly.  And by showing the conversion being done we show them what the dev needs to do when consuming values returned in juels etc.

tasks/Functions-billing/fund.js
zeuslawyer: For the subscription that has been done on line 21.

the logging can be done in LINK because the HH kit user put in the sub amount in LINK.

PR #176, opened 2023-08-28
refactor CI workflow

momentmaker:


chainchad (approved):
LGTM, but I see that we're not launching this on `main` pushes now. Is that because we have another workflow that covers running tests and lint on main?

momentmaker:
> LGTM, but I see that we're not launching this on `main` pushes now. Is that because we have another workflow that covers running tests and lint on main?

ahh sorry - good call. thanks!

chainchad (approved):

KuphJr (approved):

PR #177, opened 2023-08-29
Updated FunctionsConsumer

KuphJr:
TODO in a separate PR:
- update README
	- New args & renamed args for `functions-request` command

### DIFF
--- a/tasks/Functions-billing/accept.js
+++ b/tasks/Functions-billing/accept.js
@@ -11,7 +11,7 @@ task("functions-sub-accept", "Accepts ownership of an Functions subscription aft
 
     const accounts = await ethers.getSigners()
     if (accounts.length < 2) {
-      throw Error("This command a second wallet's private key to be made available in networks.js")
+      throw Error("This command requires a second wallet's private key to be made available in networks.js")
     }
     const accepter = accounts[1] // Second wallet.
 


Functions-request-config.js
```diff
@@ -6,6 +6,7 @@ require("@chainlink/env-enc").config()
 const Location = {
   Inline: 0,
```
zeuslawyer, line 7: remove InLine now that only the other 2 are supported?

Functions-request-config.js
```diff
@@ -6,6 +6,7 @@ require("@chainlink/env-enc").config()
...
   Remote: 1,
```
zeuslawyer, line 8: I'm starting to think that "Remote" is not a very descriptive/user friendly term -- intuitively the DON is remote too.

Should we consider "URL" instead?  Or webHosted to contra-distinguish with DONHosted?

contracts/FunctionsConsumer.sol
```diff
@@ -1,86 +1,80 @@
...
-   * @return Functions request ID
+   * @param secretsLocation Location of secrets (only Location.Remote & Location.DONHosted are supported)
+   * @param encryptedSecretsReference Reference pointing to encrypted secrets
+   * @param args String arguments that will be passed into the source code
+   * @param bytesArgs Bytes arguments that will be passed into the source code as hex strings
+   * @param subscriptionId Subscription ID used to pay for request (FunctionsConsumer contract address must first be added to the subscription)
+   * @param callbackGasLimit Maximum amount of gas used to call the client contract's `handleOracleFulfillment` method
    */
-  function executeRequest(
+  function sendRequest(
```
zeuslawyer, line 44: @KuphJr 
to address your point that people dont realise that this example contract is not exhaustive, we could have another func here called sendRequestAsCBOR() to illustrate the off chain cbor encoding of requests

contracts/@chainlink/contracts/src/v0.8/functions/dev/1_0_0/example/FunctionsClientExample.sol
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
KuphJr: This is just copied from the core repo.  It will be removed once the latest `@chainlink/contracts` package is published.

KuphJr:
I have tested everything on the staging DON & it all works as expected.

Functions-request-config.js
```diff
@@ -1,46 +1,28 @@
...
 // Configure the request by setting the fields below
 const requestConfig = {
   // Location of source code (only Inline is currently supported)
   codeLocation: Location.Inline,
+  // Location of secrets (only Remote or DONHosted is supported)
+  secretsLocation: Location.DONHosted,
   // Code language (only JavaScript is currently supported)
   codeLanguage: CodeLanguage.JavaScript,
   // String containing the source code to be executed
   source: fs.readFileSync("./calculation-example.js").toString(),
```
zeuslawyer, line 16: @KuphJr 
Wondering whether we need two examples.  The API example will convey all the necessary mental models - and a couple of comments there can help with communication that calcs without HTTP requests are feasible.

However, having to comment in and out parts of the request config is a high-friction step that can be easily overlooked and cause irritating re-runs of the CLI commands.  

This shortest path experience that builds mental models linearly can be achieved with just the API example.

Can we discuss?

Functions-request-config.js
```diff
@@ -1,46 +1,28 @@
...
+  secrets: { apiKey: process.env.COINMARKETCAP_API_KEY ?? "" },
   // ETH wallet key used to sign secrets so they cannot be accessed by a 3rd party
   walletPrivateKey: process.env["PRIVATE_KEY"],
   // Args (string only array) can be accessed within the source code with `args[index]` (ie: args[0]).
   args: ["1", "bitcoin", "btc-bitcoin", "btc", "1000000", "450"],
   // Expected type of the returned value
-  expectedReturnType: ReturnType.uint256,
+  expectedReturnType: ReturnType.string,
   // Redundant URLs which point to encrypted off-chain secrets
   secretsURLs: [],
```
zeuslawyer, line 27: worth clarifying comment with
// Redundant (un-encryped) URLs which point to encrypted off-chain secrets. Encryption will be done by functions-toolkit

Many users will wonder if they need to encrypt the URLs before putting them in this array.

contracts/FunctionsConsumer.sol
zeuslawyer: My sense is that fewer people will start with AutomatedFunctionsConsumer as it is more advanced and requires more config (higher time to satisfied feeling)

This contract will be viewed and interacted with earlier in the user journey most of the time, and also may be demo'd more.  I recommend it be here as well.  

Once request.js is done I can base a different CLI task off of that to run CBOR encoded requests if you like

contracts/FunctionsConsumer.sol
```diff
@@ -1,86 +1,80 @@
...
    * @param source JavaScript source code
-   * @param secrets Encrypted secrets payload
-   * @param args List of arguments accessible from within the source code
-   * @param subscriptionId Funtions billing subscription ID
-   * @param gasLimit Maximum amount of gas used to call the client contract's `handleOracleFulfillment` function
-   * @return Functions request ID
+   * @param secretsLocation Location of secrets (only Location.Remote & Location.DONHosted are supported)
+   * @param encryptedSecretsReference Reference pointing to encrypted secrets
+   * @param args String arguments that will be passed into the source code
+   * @param bytesArgs Bytes arguments that will be passed into the source code as hex strings
```
zeuslawyer, line 40: @param args String arguments that will be passed into the source code
@param bytesArgs Bytes arguments that will be passed into the source code as hex strings

Couple of thoughts:
1) both are optional so recommend removing "will" and just say "arguments to be passed..."
2) we may need to also consider smoothing out the language to distinguish when they'd use bytesArgs instead of just regular args

contracts/FunctionsConsumer.sol
```diff
@@ -1,86 +1,80 @@
...
-   * @param oracle New oracle address
-   */
-  function updateOracleAddress(address oracle) public onlyOwner {
-    setOracle(oracle);
-  }
-
-  function addSimulatedRequestId(address oracleAddress, bytes32 requestId) public onlyOwner {
-    addExternalRequest(oracleAddress, requestId);
+    if (s_lastRequestId != requestId) {
+      revert UnexpectedRequestID(requestId);
```
zeuslawyer, line 75: in what circumstances could this happen? how can we help end users debug this?

@aelmanaa do we have a section in the docs for common "gotchas" and debugging steps?

tasks/Functions-billing/accept.js
```diff
@@ -11,7 +11,7 @@ task("functions-sub-accept", "Accepts ownership of an Functions subscription aft
 
     const accounts = await ethers.getSigners()
     if (accounts.length < 2) {
-      throw Error("This command a second wallets's private key to be made available in networks.js")
+      throw Error("This command a second wallet's private key to be made available in networks.js")
```
zeuslawyer, line 14: nit: This command *needs*....
this was my code, so my bad!

zeuslawyer (commented):
LGTM , with a couple of comments for @aelmanaa and @KuphJr. But can be merged and we can address later so as not to block.

Functions-request-config.js
aelmanaa: @zeuslawyer Once you finish this PR, we will have to rebase the [tutorials](https://github.com/smartcontractkit/functions-hardhat-starter-kit/tree/tutorials) branch.
You can see that we only use what's necessary for each example. For instance, this is the [config](https://github.com/smartcontractkit/functions-hardhat-starter-kit/blob/tutorials/tutorials/1-simple-computation/config.js) for the simple computation tutorial.
So you can just run any custom config using `--configpath tutorials/1-simple-computation/config. For instance:

```
npx hardhat functions-request --subid REPLACE_SUBSCRIPTION_ID --contract REPLACE_CONSUMER_CONTRACT_ADDRESS --network REPLACE_NETWORK --configpath REPLACE_CONFIG_PATH
```

> @KuphJr @aelmanaa Wondering whether we need two examples. The API example will convey all the necessary mental models - and a couple of comments there can help with communication that calcs without HTTP requests are feasible.
> 
> However, having to comment in and out parts of the request config is a high-friction step that can be easily overlooked and cause irritating re-runs of the CLI commands.
> 
> This shortest path experience that builds mental models linearly can be achieved with just the API example.
> 
> **This would also reduce the amount of documentation people have to wade through**....but conscious that @aelmanaa may be burdened with updating the docs if we remove the calculation example.
> 
> Can we discuss?

@zeuslawyer Once you finish this PR, we will have to rebase the [tutorials](https://github.com/smartcontractkit/functions-hardhat-starter-kit/tree/tutorials) branch.
You can see that we only use what's necessary for each example. For instance, this is the [config](https://github.com/smartcontractkit/functions-hardhat-starter-kit/blob/tutorials/tutorials/1-simple-computation/config.js) for the simple computation tutorial.
So you can just run any custom config using `--configpath tutorials/1-simple-computation/config. For instance:

```
npx hardhat functions-request --subid REPLACE_SUBSCRIPTION_ID --contract REPLACE_CONSUMER_CONTRACT_ADDRESS --network REPLACE_NETWORK --configpath REPLACE_CONFIG_PATH
```

zeuslawyer: thanks @aelmanaa  - that makes sense since its on the /tutorials branch.  But even on that branch there is the [Functions-request-config.js](https://github.com/smartcontractkit/functions-hardhat-starter-kit/blob/tutorials/Functions-request-config.js) file which has all the commenting in and out of source files.  

And the /tutorials folder has a different source for each example. Which means in the root we really dont need ton have both the simple example and the API example.  

It adds overhead and complexity IMO!

contracts/FunctionsConsumer.sol
KuphJr: Yeah.  If you feel strongly that you want an on-demand example that uses an off-chain encoded CBOR request, I would prefer to keep that in a separate example to keep `request.js` as straightforward as possible.

We can revisit once we complete the currently scoped work for the starter kit.

Functions-request-config.js
KuphJr: If you think we should simplify and remove one of the example source code files, I am fine with that.  We can address in a separate PR.

contracts/FunctionsConsumer.sol
KuphJr: This only applies if 2 requests are in-flight.  Then this contract will only allow for receiving the latest one.  However, we could just remove this to allow for multiple in-flight responses.

I am going to remove the `if` statement since it is not necessary and just adds complexity.

contracts/FunctionsConsumer.sol
KuphJr: args & bytesArgs are always passed, but if none are provided it will just be an empty array.

Functions-request-config.js
KuphJr: They are actually no longer used as we just default to using the automatically created Gist URLs in `request.js` (which I think is way less confusing for beginners).  I removed secretsURLs to avoid confusion.

contracts/FunctionsConsumer.sol
KuphJr: For #2 I think that is beyond the scope of a code comment.  We should rely on docs for this.

zeuslawyer (approved):

PR #178, opened 2023-08-29
Update secrets management tasks.

zeuslawyer:
1. Build Off Chain Secrets and store encrypted secrets in local JSON file.
2. Upload Secrets to DON (polygonMumbai)
3. fetch and display DON-hosted secrets
4. use deleteGists in task
5. update networks.js config file to fix bug where SECOND_PRIVATE_KEY is unset


tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -8,40 +9,35 @@ task(
...
+    const secrets = {
+      // This secret is used in API-request-example.js
+      // Get free API key from https://pro.coinmarketcap.com/
+      apiKey: process.env.CMC_API_KEY,
+    }
+
+    console.log("\nEncrypting secrets and writing to JSON file...")
+    const encryptedSecretsObj = await secretsManager.buildEncryptedSecrets(secrets)
+
+    const outputfile = taskArgs.output ? taskArgs.output : "offchain-encyrpted-secrets.json"
```
KuphJr, line 39: fix filename typo

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -8,40 +9,35 @@ task(
...
-    `${__dirname}/../../Functions-request-config.js`,
-    types.string
-  )
+  .addOptionalParam("output", "Output JSON file name (defaults to offchain-secrets.json)")
   .setAction(async (taskArgs) => {
     if (network.name === "hardhat") {
       throw Error("This command cannot be used on a local hardhat chain.")
     }
 
-    const requestConfig = require(path.isAbsolute(taskArgs.configpath)
```
KuphJr, line 23: This `requestConfig` should contain the secrets we want to use.  It shouldn't be removed.

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -8,40 +9,35 @@ task(
...
 
-    const offchainSecrets = await generateOffchainSecrets(
-      requestConfig,
-      process.env.PRIVATE_KEY,
-      DONPublicKey,
-      nodeAddresses,
-      perNodePublicKeys
-    )
+    // TODO @dev use the env-enc package to encrypt secrets and write into a .env.enc file.
+    const secrets = {
```
KuphJr, line 30: use secrets from `requestConfig`

tasks/Functions-client/buildOffchainSecrets.js
```diff
@@ -8,40 +9,35 @@ task(
   "functions-build-offchain-secrets",
   "Builds an off-chain secrets object that can be uploaded and referenced via URL"
 )
-  .addOptionalParam("output", "Output file name (defaults to offchain-secrets.json)")
```
KuphJr, line 11: You can add a default output file name here instead of using a ternary operator below.

zeuslawyer: changed to ...(defaults to offchain-encrypted-secrets.json)")

KuphJr (commented):
LGTM. I'll try to debug the gateway issues so you can thoroughly test each one of these commands, then take the PR out of draft.

KuphJr (changes requested):
It looks like when you added the second private key for subscription transferring, it breaks the tooling unless the user provides the SECOND_PRIVATE_KEY.  Can we fix this?

networks.js
```diff
@@ -13,55 +13,68 @@ const SHARED_DON_PUBLIC_KEY =
...
     confirmations: DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS,
     nativeCurrencySymbol: "ETH",
     linkToken: "0x779877A7B0D9E8603169DdbD7836e478b4624789",
     linkPriceFeed: "0x42585eD362B3f1BCa95c640FdFf35Ef899212734",
-    functionsOracleProxy: "0x649a2C205BE7A3d5e99206CEEFF30c794f0E31EC",
-    functionsBillingRegistryProxy: "0x3c79f56407DCB9dc9b852D139a317246f43750Cc",
+    functionsRouter: "", // TODO @zeuslawyer
+    donId: "", // TODO @zeuslawyer
+    gatewayUrls: "", // TODO @zeuslawyer
     functionsPublicKey: SHARED_DON_PUBLIC_KEY,
```
KuphJr, line 48: We can also entirely remove SHARED_DON_PUBLIC_KEY from this file. It is no longer ever needed locally as it will always be pulled from on-chain.

PR #179, opened 2023-09-04
Autofunc

zeuslawyer:


listen.js
```diff
@@ -0,0 +1,33 @@
+const { networks } = require("./networks")
```
KuphJr, line 1: Lets put this file in the `scripts` folder.  If we want we can add `nodemon`, but I think just putting a command in package.json so you can do something like `npm run listen` would be sufficient too.

tasks/Functions-client/setAutoRequest.js
```diff
@@ -1,15 +1,21 @@
...
 const process = require("process")
 
-task("functions-set-auto-request", "Updates the Functions request in a deployed AutomatedFunctionsConsumer contract")
+task(
+  "functions-set-auto-request",
+  "sets the CBOR-encoded Functions request in a deployed AutomatedFunctionsConsumer contract"
+)
   .addParam("contract", "Address of the client contract")
   .addParam("subid", "Billing subscription ID used to pay for Functions requests", undefined, types.int)
+  .addParam(
```
KuphJr, line 15: This should be an optional param in case the user has specified `Location.remote` in Functions-request-config.js

tasks/Functions-client/setAutoRequest.js
```diff
@@ -44,92 +50,74 @@ const setAutoRequest = async (contract, taskArgs) => {
...
   const unvalidatedRequestConfig = require(path.isAbsolute(taskArgs.configpath)
     ? taskArgs.configpath
     : path.join(process.cwd(), taskArgs.configpath))
+
   const requestConfig = getRequestConfig(unvalidatedRequestConfig)
 
-  // doGistCleanup indicates if an encrypted secrets Gist was created automatically and should be cleaned up by the user after use
-  let doGistCleanup = !(requestConfig.secretsURLs && requestConfig.secretsURLs.length > 0)
-  const request = await generateRequest(requestConfig, taskArgs)
+  console.log("\nEncrypting secrets and uploading to DON...")
```
KuphJr, line 64: Do you also want to support remote secrets or do you prefer to exclusively use DON hosted?
While it might be nice to give users both options, I do see risk that users won't remember to delete their Gists...

If we choose not to support `Location.Remote` in this script, then we should give some sort of error if the user specifies `Location.Remote` in Functions-request-config.js

tasks/Functions-client/setAutoRequest.js
```diff
@@ -44,92 +50,74 @@ const setAutoRequest = async (contract, taskArgs) => {
...
+  const functionsRouterAddress = networks[network.name]["functionsRouter"]
+  const donId = networks[network.name]["donId"]
+  const signer = await ethers.getSigner()
```
KuphJr, line 68: Check if the user has actually specified secrets (secrets && Object.keys(secrets) > 0) before doing the secrets workflow.

tasks/Functions-client/setAutoRequest.js
```diff
@@ -44,92 +50,74 @@ const setAutoRequest = async (contract, taskArgs) => {
...
-    /* new request, continue */
+  const { version, success } = await secretsManager.uploadEncryptedSecretsToDON({
+    encryptedSecretsHexstring: encryptedSecretsObj.encryptedSecrets,
+    gatewayUrls: networks[network.name]["gatewayUrls"],
+    storageSlotId: slotId,
+    minutesUntilExpiration,
+  })
+
+  if (!success) {
+    throw Error("\nFailed tp upload encrypted secrets to DON.")
```
KuphJr, line 88: Note that `uploadEncryptedSecretsToDON` will throw a descriptive error if it fails for all nodes.  success == false just indicates that one of the nodes failed.  Not quite sure how best to communicate this to users though...

tasks/Functions-client/deployAutoClient.js
```diff
@@ -36,42 +37,48 @@ task("functions-deploy-auto-client", "Deploys the AutomatedFunctionsConsumer con
       throw Error("Gas limit must be less than or equal to 300,000")
```
KuphJr, line 37: Remove this gas limit check.  Instead, use `subManager.estimateFunctionsRequestCost`. This is because it is possible for certain subscriptions which we manually allow to use more than 300k gas.  This logic is all verified by estimateCost.  See here: https://github.com/smartcontractkit/functions-hardhat-starter-kit/blob/develop/tasks/Functions-client/request.js#L105

tasks/Functions-client/deployAutoClient.js
```diff
@@ -36,42 +37,48 @@ task("functions-deploy-auto-client", "Deploys the AutomatedFunctionsConsumer con
...
+    const consumerAddress = autoClientContract.address
 
-    taskArgs.contract = autoClientContract.address
+    const sm = new SubscriptionManager({ signer, linkTokenAddress, functionsRouterAddress })
+    await sm.initialize()
 
-    await setAutoRequest(autoClientContract.address, taskArgs)
+    console.log(`\nAdding ${consumerAddress} to subscription ${subId}...`)
+    const addConsumerTx = await sm.addConsumer({ subId, consumerAddress, txOptions })
+    console.log(`\nAdded consumer contract ${consumerAddress} in Tx: ${addConsumerTx.transactionHash}`)
```
KuphJr, line 73: I am wondering if it makes sense to set the request here too upon deployment.  Basically, instead of users needing to deploy and then run set-auto-request, they could just deploy and the request would be set automatically.  This could be accomplished by building the CBOR bytes in this file and adding the requestCBOR as a constructor arg.

README.md
```diff
@@ -269,18 +269,34 @@ URLs which host secrets must be available every time a request is executed by DO
...
-2. Deploy the _AutomationFunctionsConsumer_ client contract by running:<br>`npx hardhat functions-deploy-auto-client --network network_name_here --subid subscription_id_number_here --interval time_between_requests_here --verify true`<br>**Note**: Make sure `<explorer>_API_KEY` environment variable is set. API keys for these services are freely available to anyone who creates an EtherScan, PolygonScan or SnowTrace account.<br><br>
-3. Register the contract for upkeep via the Chainlink Automation web app here: [https://automation.chain.link/](https://automation.chain.link/)
+
+2. Deploy the _AutomationFunctionsConsumer_ client contract by running:<br>`npx hardhat functions-deploy-auto-client --network network_name_here --subid subscription_id_number_here --interval time_between_requests_here --verify true`<br>**Note**: Make sure `<blockexplorer>_API_KEY` environment variable is se - this CLI command verifies the contract using this API KEY. API keys for these services are freely available to anyone who creates an EtherScan, PolygonScan or SnowTrace account.<br><br>
```
KuphJr, line 273: typo "se -" -> "set"

KuphJr: Reread this line to make sure it makes sense.

README.md
```diff
@@ -269,18 +269,34 @@ URLs which host secrets must be available every time a request is executed by DO
...
+
+3. Encode the request parameters into CBOR and store it on chain with `npx hardhat functions-set-auto-request --network network_name_here  --subid subscription_id_number_here --interval time_between_requests_here --contract 0x_contract_address`.  You can now manually check that your on-chain requests work if you try the manual debugging step #4 or you can go ahead and run Chainlink Automations on your contract by following the next step.  
+
+> ⚠️⚠️⚠️ Keep in mind that this task sets DON-Hosted secrets and expires those secrets after 10 minutes.  If you see error bytes returned to your client which decode to secrets not being found, you should run this functions-set-auto-request command or the functions-upload-secrets-don to refresh the DON-hosted secrets for that slotId.
```
KuphJr, line 277: Let's add a ttl as a CLI argument.  Make the default sometime longer like an hour or a day.  I am not very concerned about longer lived secrets if they are DON hosted.

tasks/Functions-client/deployAutoClient.js
KuphJr: Else... Maybe we do require users to run `set-auto-request`, but then we don't need to pass the subId, gasLimit or interval as constructor args.

README.md
```diff
@@ -269,18 +269,34 @@ URLs which host secrets must be available every time a request is executed by DO
...
    - Be sure to set the `Gas limit` for the _performUpkeep_ function to a high enough value. The recommended value is 1,000,000.
    - Find further documentation for working with Chainlink Automation here: [https://docs.chain.link/chainlink-automation/introduction](https://docs.chain.link/chainlink-automation/introduction)
 
-Once the contract is registered for upkeep, check the latest response or error with the commands `npx hardhat functions-read --network network_name_here --contract contract_address_here`.
+Once the contract is registered for upkeep, check the latest response or error with the commands `npx hardhat functions-read --network network_name_here --contract contract_address_here`.  Note that you can also pause and unpause your Automation Upkeep using the web app's UI.
+
+4. For debugging, use the command `npx hardhat functions-check-upkeep --network network_name_here --contract contract_address_here` to see if Automation needs to call _performUpkeep_. If this call returns `false` then the upkeek interval has not yet passed and `performUpkeep` will not execute.
```
KuphJr, line 285: typo "upkeek"

README.md
```diff
@@ -269,18 +269,34 @@ URLs which host secrets must be available every time a request is executed by DO
...
+
+You can also attach a listener to a Subscription ID by updating the `network` and `subId` variables in `listen.js` and then running `node listen.js` from the repo root. 
+
+Once the listener is running (and after each update interval), you can also manually trigger a request by using the command `npx hardhat functions-perform-upkeep --network network_name_here --contract contract_address_here`. If your DON hosted secrets have not been expired and the JS source does not throw errors during execution, you will see the result of the computation printed to your console.
```
KuphJr, line 289: This line is just a bit confusing as the listener does not need to be running in order to perform a manual upkeep.
TBH, I think we could just remove this line altogether as functions-perform-upkeep should only ever need to be used if Automation is broken...
In fact, functions-perform-upkeep likely won't even work because the Automation nodes will likely perform the upkeep before a user even has time to run the CLI command.

tasks/Functions-client/deployAutoClient.js
zeuslawyer: Have kept this separate after a bit of deliberation as i feel that combining the two steps into the deploy step will "hide" and important step needed for the dev to understand the lifecylcle/flow.

tasks/Functions-client/setAutoRequest.js
zeuslawyer: I think its best to just illustrate DON Hosted as that is the more secure and preferred approach.

tasks/Functions-client/setAutoRequest.js
zeuslawyer: Great catch!

tasks/Functions-client/setAutoRequest.js
zeuslawyer: Is this still accurate given the changes we noted when there are two gateway URLs being passed? I got a descriptive error when the second URL threw an error (which indicates first URL was fine, and success would == false)

https://chainlink-core.slack.com/archives/C04DBHV754K/p1694036838760739?thread_ts=1694036015.102869&cid=C04DBHV754K

zeuslawyer: I checked the stack and the error I linked to arises from secretsManager.uploadEncryptedSecretsToDON() which means that even that it seems to be throwing an error even if one node succeeded

KuphJr: Gotcha.  Bolek & I will discuss this and I may need to handle this better in the toolkit.

README.md
```diff
@@ -270,25 +270,21 @@ Chainlink Functions can be used with Chainlink Automation in order to automatica
...
 
 3. Encode the request parameters into CBOR and store it on chain with `npx hardhat functions-set-auto-request --network network_name_here  --subid subscription_id_number_here --interval time_between_requests_here --contract 0x_contract_address`.  You can now manually check that your on-chain requests work if you try the manual debugging step #4 or you can go ahead and run Chainlink Automations on your contract by following the next step.  
 
-> ⚠️⚠️⚠️ Keep in mind that this task sets DON-Hosted secrets and expires those secrets after 10 minutes.  If you see error bytes returned to your client which decode to secrets not being found, you should run this functions-set-auto-request command or the functions-upload-secrets-don to refresh the DON-hosted secrets for that slotId.
+> ⚠️Keep in mind that this task sets DON-Hosted secrets and expires those secrets after 10 minutes.  If you see error bytes returned to your client which decode to secrets not being found, you should run this functions-set-auto-request command or the functions-upload-secrets-don to refresh the DON-hosted secrets for that slotId.
```
KuphJr, line 277: Update this with mention of the TTL CLI arg.

Also missing a space before keep.

README.md
```diff
@@ -270,25 +270,21 @@ Chainlink Functions can be used with Chainlink Automation in order to automatica
...
-
-Once the listener is running (and after each update interval), you can also manually trigger a request by using the command `npx hardhat functions-perform-upkeep --network network_name_here --contract contract_address_here`. If your DON hosted secrets have not been expired and the JS source does not throw errors during execution, you will see the result of the computation printed to your console.
-
+4. For debugging, use the command `npx hardhat functions-check-upkeep --network network_name_here --contract contract_address_here` to see if Automation needs to call _performUpkeep_. If this call returns `false` then the upkeep interval has not yet passed and `performUpkeep` will not execute.
 
+You can also attach a listener to a Subscription ID by updating the `network` and `subId` variables in `listen.js` and then running `node listen.js` from the repo root. To do this open a new terminal or split terminal and run `npm run listen`. This script uses nodemon which restarts the script when you save files or when the listener returns a result.
```
KuphJr, line 287: make sure this line is up-to-date

tasks/Functions-billing/transfer.js
```diff
@@ -13,7 +13,7 @@ task("functions-sub-transfer", "Request ownership of an Functions subscription b
       )
     }
 
-    const subId = taskArgs.subid
+    const subscriptionId = taskArgs.subid
```
KuphJr, line 16: could do parseInt here

tasks/Functions-client/deployAutoClient.js
```diff
@@ -33,42 +32,48 @@ task("functions-deploy-auto-client", "Deploys the AutomatedFunctionsConsumer con
...
+    const callbackGasLimit = taskArgs.gaslimit
 
+    // Initialize SubscriptionManager
+    const subManager = new SubscriptionManager({ signer, linkTokenAddress, functionsRouterAddress })
+    await subManager.initialize()
+
+    // Validate callbackGasLimit
+    const { gasPrice } = await hre.ethers.provider.getFeeData()
+    const gasPriceGwei = BigInt(Math.ceil(hre.ethers.utils.formatUnits(gasPrice, "gwei").toString()))
+    _ = await subManager.estimateFunctionsRequestCost({
```
KuphJr, line 55: remove unused var. Just `await subManager...`

tasks/Functions-client/setAutoRequest.js
```diff
@@ -61,39 +90,49 @@ const setAutoRequest = async (contract, taskArgs) => {
...
-  const minutesUntilExpiration = 10 // Minimum 5 minutes supported.
-
-  const { version, success } = await secretsManager.uploadEncryptedSecretsToDON({
-    encryptedSecretsHexstring: encryptedSecretsObj.encryptedSecrets,
-    gatewayUrls: networks[network.name]["gatewayUrls"],
-    storageSlotId: slotId,
-    minutesUntilExpiration,
-  })
+  if (!requestConfig.secrets || Object.keys(requestConfig.secrets).length === 0) {
+    throw Error("\nThis task requires a secrets object in request config. None found")
```
KuphJr, line 94: What if the user doesn't want to use secrets?  I'm not sure we should force them to use secrets here.

KuphJr (commented):
Few more considerations.  LMK when and I'll give a final review.

tasks/Functions-client/setAutoRequest.js
KuphJr: In which case, slotId could be optional too.  We just need to verify it is present if the user has specified secrets.

README.md
```diff
@@ -269,18 +269,30 @@ URLs which host secrets must be available every time a request is executed by DO
...
 
-For debugging, use the command `npx hardhat functions-check-upkeep --network network_name_here --contract contract_address_here` to see if Automation needs to call _performUpkeep_.
-To manually trigger a request, use the command `npx hardhat functions-perform-upkeep --network network_name_here --contract contract_address_here`.
+4. For debugging, use the command `npx hardhat functions-check-upkeep --network network_name_here --contract contract_address_here` to see if Automation needs to call _performUpkeep_. If this call returns `false` then the upkeep interval has not yet passed and `performUpkeep` will not execute.
+
+You can also attach a listener to a Subscription ID by updating the  `subId` variable in `listen.js` and then running `npm run listen --network your_network_name` from the repo root. To do this open a new terminal or split terminal and run `npm run listen`. This script uses nodemon which restarts the script when you save files or when the listener returns a result.
```
KuphJr, line 288: nit: give the path to `listen.js` within `/scripts`

KuphJr (approved):
Just make sure you do a manual test of everything before merging.

PR #181, opened 2023-09-11
localFunctionsTestnet support

KuphJr:
- Use NPM package `simulateScript` for simulating JavaScript execution
- Added script to start local Functions testnet
- Support local Functions testnet for all existing commands
- Removed old simulate command
- Removed unused code

TODO in a separate PR:
- Update README

Also, all commands should be manually tested again to ensure no breaking changes.


zeuslawyer:
Some comments in this thread: https://chainlink-core.slack.com/archives/D03C46WBERG/p1694468257671979

tasks/Functions-client/deployAutoClient.js
```diff
@@ -77,7 +71,12 @@ task("functions-deploy-auto-client", "Deploys the AutomatedFunctionsConsumer con
     console.log(`\nAdded consumer contract ${consumerAddress} in Tx: ${addConsumerTx.transactionHash}`)
 
     const verifyContract = taskArgs.verify
-    if (verifyContract && !!networks[network.name].verifyApiKey && networks[network.name].verifyApiKey !== "UNSET") {
+    if (
+      network.name !== "localFunctionsTestnet" &&
```
zeuslawyer, line 75: This logic needs to be separated out otherwise the else logic on line 50 will run for the local testnet and tell the user that they have not set the API keys (even if they have)

The situation where `network.name == "localFunctionsTestnet" ` should just silently skip verification, so before the assignment of taskargs.verify we could just do
```
if (network.name === "localFunctionsTestnet) {
  return 
}
```

contracts/AutomatedFunctionsConsumer.sol
```diff
@@ -44,7 +44,7 @@ contract AutomatedFunctionsConsumer is FunctionsClient, ConfirmedOwner, Automati
   ) FunctionsClient(router) ConfirmedOwner(msg.sender) {
     donId = _donId;
     s_fulfillGasLimit = _fulfillGasLimit;
-    s_lastUpkeepTimeStamp = block.timestamp;
+    s_lastUpkeepTimeStamp = 0;
```
zeuslawyer, line 47: @KuphJr 
Is this only so that it works on the testnet? does the testnet ganache not come with its own starting timestamp || 0? this deviates a bit from our docs etc and may be confusing but if unavoidable, is ok.

scripts/startLocalFunctionsTestnet.js
```diff
@@ -0,0 +1,54 @@
+const process = require("process")
+const path = require("path")
+const fs = require("fs")
+const { startLocalFunctionsTestnet } = require("@chainlink/functions-toolkit")
+const { utils, Wallet } = require("ethers")
+// Loads environment variables from .env.enc file (if it exists)
+require("@chainlink/env-enc").config("../.env.enc")
+;(async () => {
```
zeuslawyer, line 8: nit: there is a semi colon here at the start?

scripts/startLocalFunctionsTestnet.js
```diff
@@ -0,0 +1,54 @@
...
+  const requestConfigPath = path.join(process.cwd(), "Functions-request-config.js") // @dev Update this to point to your desired request config file
```
zeuslawyer, line 9: may be worth console logging 
console.log(`\nRequest config used is : ${ path.join(process.cwd(), "Functions-request-config.js")... }`

so devs can be left visual clues in case they get unexpected functions results

scripts/startLocalFunctionsTestnet.js
```diff
@@ -0,0 +1,54 @@
...
+        debug: false,
+        verbose: false,
+        quiet: true, // Set this to `false` to see logs from the local testnet
+      },
+    } // Ganache server options (optional)
+  )
+
+  console.log("\nFunctionsRouter Contract Address:", localFunctionsTestnetInfo.functionsRouterContract.address)
+  console.log("DON ID:".padStart(33, " "), localFunctionsTestnetInfo.donId)
+  console.log("Mock LINK Token Contract Address:", localFunctionsTestnetInfo.linkTokenContract.address, "\n")
```
zeuslawyer, line 24: havent used this in a while, so worth checking but may be a more elegant way to log this data without needing padding
https://developer.mozilla.org/en-US/docs/Web/API/console/table

scripts/startLocalFunctionsTestnet.js
```diff
@@ -0,0 +1,54 @@
...
+
+  // Fund wallets with ETH and LINK
+  const addressToFund = new Wallet(process.env["PRIVATE_KEY"]).address
+  await localFunctionsTestnetInfo.getFunds(addressToFund, {
```
zeuslawyer, line 28: just realised that getFunds is a misleading/confusing name.  For consistency can we make this add funds?  I realise this should be in the toolkit repo! sorry I missed it there!

tasks/Functions-client/readResultAndError.js
```diff
@@ -1,4 +1,4 @@
-const { getDecodedResultLog } = require("../../FunctionsSandboxLibrary")
+const { decodeResult } = require("@chainlink/functions-toolkit")
```
zeuslawyer, line 1: good pickup - thanks!

tasks/Functions-client/request.js
```diff
@@ -128,7 +122,11 @@ task("functions-request", "Initiates an on-demand request from a Functions consu
     // Handle encrypted secrets
     let encryptedSecretsReference = []
     let gistUrl
```
zeuslawyer, line 124: gistUrl not used anymore

tasks/simulateScript.js
```diff
@@ -0,0 +1,31 @@
+const { simulateScript, decodeResult } = require("@chainlink/functions-toolkit")
+const path = require("path")
+const process = require("process")
+
+task("functions-simulate-script", "Initiates an on-demand request from a Functions consumer contract")
```
zeuslawyer, line 5: task description needs to be "Locally simulates the execution of your custom source code"

zeuslawyer (approved):
LGTM - just a couple of comments for you to consider.

contracts/AutomatedFunctionsConsumer.sol
KuphJr: This is because I want to be able to trigger an upkeep immediately after I deploy. I don't want to wait for the full time interval before I am able to call `performUpkeep` (or have the Automation nodes call it)

scripts/startLocalFunctionsTestnet.js
KuphJr: It is a JavaScript requirement for the edge case have a require statement before a `(async () => { ... })()` statement.

There are very few cases where semicolons are actually required in JS, but this is one of them.

scripts/startLocalFunctionsTestnet.js
KuphJr: I forgot about console.table. I have never had a reason to use it up until now, but it does make sense here. Thanks!

scripts/startLocalFunctionsTestnet.js
KuphJr: Hmmmm I am not sure `addFunds` is the best name either. I'll merge this as-is for now, but maybe we should think of a better name.

PR #182, opened 2023-09-14
Zp readme 2

zeuslawyer:


README.md
```diff
@@ -94,12 +94,23 @@ Install **both** of the following:
...
-   <br>**Note**: Ensure your wallet has a sufficient LINK balance before running this command. Testnet LINK can be obtained at <a href="https://faucets.chain.link/">faucets.chain.link</a>.<br><br>
-9. Make an on-chain request by running:<br>`npx hardhat functions-request --network network_name_here --contract 0xDeployed_consumer_contract_address_here --subid subscription_id_number_here`
+8. Create and fund a new Functions billing subscription, and then authorize your deployed consumer contract to use that subscription by running:<br> `npx hardhat functions-sub-create --network network_name_here --amount LINK_funding_amount_here --contract 0xDeployed_client_contract_address_here`<br><br>**Note**: Ensure your wallet has a sufficient LINK balance before running this command. Testnet LINK can be obtained at <a href="https://faucets.chain.link/">faucets.chain.link</a>.
```
KuphJr, line 101: I am wondering if, for the quickstart, we should just instruct users on using the frontend to create their subscription as they will already need to use the frontend to accept the terms of service/add themselves to the allowlist.
I think the majority of users will just use the frontend for subscription management anyway.

KuphJr: I don't think the instructions need to be detailed since this step is so straightforward. Maybe just say something like "create a subscription via the Chainlink Functions website" & give them a link.

README.md
```diff
@@ -94,12 +94,23 @@ Install **both** of the following:
...
+   **Note**: also make a note of your subscription Id as you will need it for most commands. <br><br>
+9. Make an on-chain request by running:<br>`npx hardhat functions-request --network network_name_here --contract 0xDeployed_client_contract_address_here --subid subscription_id_number_here`. You will see a confirmation request, so hit `Y` and press enter. Once the request is fulfilled the console will show the response (decoded into the relevant return type) from the execution of your custom JS script.
+
+10. You can also query the response that was stored in your Functions Consumer contract by runnning `npx hardhat functions-read --contract consumer_contract_address --network  your_network_name`
```
KuphJr, line 105: Double check that this works. I might have screwed something up locally, but I think I had an error here...

README.md
KuphJr: Make sure to use "consumer", not "client"

README.md
```diff
@@ -311,11 +322,11 @@ Chainlink Functions can be used with [Chainlink Automation](https://docs.chain.l
...
 
-You can also attach a listener to a Subscription ID by updating the `subId` variable in `/scripts/listen.js`, and then running `npm run listen --network your_network_name` from the repo root. To do this open a new terminal or split terminal and run `npm run listen`. This script uses nodemon which restarts the script when you save files or when the listener returns a result.
+You can also attach a listener to a Subscription ID by updating the `subId` variable in `/scripts/listen.js`, and then running `npm run listen --network your_network_name` from the repo root. **Note** Do this in a new terminal or split terminal so that it can keep listening as you develop. This script uses nodemon which restarts the script when you save files or when the listener returns a result.
```
KuphJr, line 329: We should test this too. IIRC I might have noticed an issue here but didn't have time to re-test

KuphJr (commented):
Thanks for the update! Couple of comments, but looks good!

README.md
zeuslawyer: works as at my current version which is one behind the lint fix pushed in 4e165605ba124c398d8b99d2d9a95e6d2fc39b8f

README.md
zeuslawyer: works as at my current version which is one behind the lint fix pushed in 4e165605ba124c398d8b99d2d9a95e6d2fc39b8f

PR #190, opened 2023-10-27
responseListener.listenForResponseFromTransaction

KuphJr:
Use new `responseListener.listenForResponseFromTransaction` method from the latest Functions Toolkit release

### DIFF
--- a/networks.js
+++ b/networks.js
@@ -101,7 +101,7 @@ const networks = {
     accounts,
     verifyApiKey: process.env.POLYGONSCAN_API_KEY || "UNSET",
     chainId: 80001,
-    confirmations: DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS * 4,
+    confirmations: 10,
     nativeCurrencySymbol: "MATIC",
     linkToken: "0x326C977E6efc84E512bB9C30f76E30c160eD06FB",
     linkPriceFeed: "0x12162c3E810393dEC01362aBf156D7ecf6159528", // LINK/MATIC
--- a/networks.js
+++ b/networks.js
@@ -101,7 +101,7 @@ const networks = {
     accounts,
     verifyApiKey: process.env.POLYGONSCAN_API_KEY || "UNSET",
     chainId: 80001,
-    confirmations: DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS,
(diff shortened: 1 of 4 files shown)


tasks/Functions-consumer/request.js
```diff
@@ -206,15 +206,14 @@ task("functions-request", "Initiates an on-demand request from a Functions consu
     }
 
     // Listen for fulfillment
-    const requestId = requestTxReceipt.events[2].args.id
     spinner.start(
-      `Request ${requestId} has been initiated. Waiting for fulfillment from the Decentralized Oracle Network...\n`
+      `Request ${requestTxReceipt.events[2].args.id} has been initiated. Waiting for fulfillment from the Decentralized Oracle Network...\n`
```
bolekk, line 210: I changed other examples to print a transaction hash here and the request ID only after fulfillment. Maybe do the same here?

networks.js
```diff
@@ -101,7 +101,7 @@ const networks = {
     accounts,
     verifyApiKey: process.env.POLYGONSCAN_API_KEY || "UNSET",
     chainId: 80001,
-    confirmations: DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS,
+    confirmations: DEFAULT_VERIFICATION_BLOCK_CONFIRMATIONS * 4,
```
bolekk, line 104: That looks a little weird. Does it actually hurt to keep it at DEFAULT?

zeuslawyer: do we want to have network specific defaults?

tasks/Functions-consumer/request.js
zeuslawyer: Is there informational value in having both? something along the lines of:

`RequestId ${reqId} is waiting for fulfillment from the Decentralized Oracle Network.  Track this fulfillment by its Tx hash ${txHash}. `

zeuslawyer (commented):
LGTM

pinebit (approved):

networks.js
KuphJr: I'll just set this to 10 blocks. I've noticed that it can be a problem if we don't wait long enough before verifying contracts.

bolekk: 10 blocks on Sepolia is 2 minutes, way too slow IMO

PR #195, opened 2023-11-21
Use try-catch for sending Functions requests

KuphJr:
- This corrects a potential bug if a Functions subscription lacks the funds to send a request
- Previously, it was possible for `performUpkeep` to revert if the Functions subscription lacked funds. This meant the last upkeep time was never updated, so Automation kept retrying the upkeep and could spend all the funds in the Automation subscription.
- The `try-catch` prevents Automation from attempting repeated failed upkeeps since the last upkeep time is always updated even if the Functions request fails instead of the entire transaction reverting

### DIFF
--- a/contracts/AutomatedFunctionsConsumer.sol
+++ b/contracts/AutomatedFunctionsConsumer.sol
@@ -26,9 +26,12 @@ contract AutomatedFunctionsConsumer is FunctionsClient, ConfirmedOwner, Automati
   uint256 public s_updateInterval;
   uint256 public s_lastUpkeepTimeStamp;
   uint256 public s_upkeepCounter;
+  uint256 public s_requestCounter;
   uint256 public s_responseCounter;
 
   event OCRResponse(bytes32 indexed requestId, bytes result, bytes err);
+  event RequestRevertedWithErrorMsg(string reason);
+  event RequestRevertedWithoutErrorMsg(bytes data);
 
   /**
    * @notice Executes once when a contract is created to initialize state variables
@@ -86,8 +89,23 @@ contract AutomatedFunctionsConsumer is FunctionsClient, ConfirmedOwner, Automati
     s_lastUpkeepTimeStamp = block.timestamp;
     s_upkeepCounter = s_upkeepCounter + 1;
 
-    bytes32 requestId = _sendRequest(s_requestCBOR, s_subscriptionId, s_fulfillGasLimit, donId);
-    s_lastRequestId = requestId;
+    try
+      i_router.sendRequest(
+        s_subscriptionId,
+        s_requestCBOR,
+        FunctionsRequest.REQUEST_DATA_VERSION,
+        s_fulfillGasLimit,
+        donId
+      )
+    returns (bytes32 requestId) {
+      s_requestCounter = s_requestCounter + 1;
+      s_lastRequestId = requestId;
+      emit RequestSent(requestId);
+    } catch Error(string memory reason) {
+      emit RequestRevertedWithErrorMsg(reason);
+    } catch (bytes memory data) {
+      emit RequestRevertedWithoutErrorMsg(data);
+    }
   }
 
   /**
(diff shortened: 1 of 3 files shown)


tasks/Functions-consumer/performManualUpkeep.js
```diff
@@ -16,22 +16,42 @@ task("functions-perform-upkeep", "Manually call performUpkeep in an Automation c
     // Call performUpkeep
     const performData = taskArgs.data ?? []
 
+    const autoConsumerContractFactory = await ethers.getContractFactory("AutomatedFunctionsConsumer")
+    const autoConsumerContract = await autoConsumerContractFactory.attach(taskArgs.contract)
+
+    const checkUpkeep = await autoConsumerContract.checkUpkeep(performData)
```
zeuslawyer, line 22: Possibly a nit:  Given that the goal is to make sure that we do not re run automation when the sub has no balance could we not just
1) performUpkeep anyway and handle a reviert if checkUpkeep returns false (this logic is now in performupkeep)
2) console log the error (with string or without) if the functions request fails

Do we need to call checkUpkeep separately here?

zeuslawyer (commented):
Minor question

tasks/Functions-consumer/performManualUpkeep.js
KuphJr: I updated the console.log message slightly, although I do think how it is currently implemented is how I intended.

contracts/AutomatedFunctionsConsumer.sol
```diff
@@ -86,8 +89,23 @@ contract AutomatedFunctionsConsumer is FunctionsClient, ConfirmedOwner, Automati
     s_lastUpkeepTimeStamp = block.timestamp;
     s_upkeepCounter = s_upkeepCounter + 1;
 
-    bytes32 requestId = _sendRequest(s_requestCBOR, s_subscriptionId, s_fulfillGasLimit, donId);
-    s_lastRequestId = requestId;
+    try
+      i_router.sendRequest(
```
bolekk, line 93: Why not _sendRequest any more?

contracts/AutomatedFunctionsConsumer.sol
```diff
@@ -26,9 +26,12 @@ contract AutomatedFunctionsConsumer is FunctionsClient, ConfirmedOwner, Automati
   uint256 public s_updateInterval;
   uint256 public s_lastUpkeepTimeStamp;
   uint256 public s_upkeepCounter;
+  uint256 public s_requestCounter;
```
bolekk, line 29: What is this counter for? Just to check how many responses vs requests we had so far?

tasks/Functions-consumer/performManualUpkeep.js
```diff
@@ -16,22 +16,42 @@ task("functions-perform-upkeep", "Manually call performUpkeep in an Automation c
...
+    if (!checkUpkeep.upkeepNeeded) {
+      console.log("\ncheckUpkeep returned false. Upkeep was not performed.")
+      return
```
bolekk, line 25: This might be a little annoying if somebody wants to force a performUpkeep but now needs to wait for s_updateInterval to pass. How is this command expected to be used? Maybe put calling checkUpkeep under a flag?

tasks/Functions-consumer/performManualUpkeep.js
zeuslawyer: sorry @KuphJr  i commented over too many lines! I was wondering whether line 22 could just go straight to performUpkeep (and rely on its internal checkupkeep call).

contracts/AutomatedFunctionsConsumer.sol
KuphJr: `try-catch` can only be used for external function calls

contracts/AutomatedFunctionsConsumer.sol
KuphJr: Yeah. Not necessary, but might be useful so I added it. I like to use it for heartbeats.

tasks/Functions-consumer/performManualUpkeep.js
KuphJr: Noted. I can just remove this check and allow `performUpkeep` to revert if `checkUpkeep` returns false.

contracts/AutomatedFunctionsConsumer.sol
bolekk: TIL, thanks!

pinebit (approved):

PR #206, opened 2024-02-14
base sepolia staging

KuphJr:


uploadSecrets.js
zeuslawyer, line 1: @KuphJr this task is already there? `functions-upload-secrets-don`

zeuslawyer:
Done in https://github.com/smartcontractkit/functions-hardhat-starter-kit/pull/209

PR #207, opened 2024-03-07
Remove hardhat-etherscan as its been deprecated in

zeuslawyer:
favour of hardhat-verify.

hardhat-toolbox includes hh-verify.

Add arb sepolia to custom networks for verification functionality.

### DIFF
--- a/hardhat.config.js
+++ b/hardhat.config.js
@@ -54,6 +54,16 @@ module.exports = {
       arbitrum: networks.arbitrum.verifyApiKey,
       arbitrumSepolia: networks.arbitrumSepolia.verifyApiKey,
     },
+    customChains: [
+      {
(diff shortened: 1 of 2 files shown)


KuphJr:
Nice! Will this break the `--verify` flat when running the `npx hardhat functions-deploy-consumer` command or does that still work as expected?

KuphJr (approved):

zeuslawyer:
> Nice! Will this break the `--verify` flat when running the `npx hardhat functions-deploy-consumer` command or does that still work as expected?

nope-- WAI with CLI commands. See for example this verified contract on Op Sep

https://sepolia-optimistic.etherscan.io/address/0x8eb0561Ace3168930Cc698885A9e0904280cd77b#code

